PR #5, opened 2018-08-28
Fixed the subscription race condition

dfee:
I ended up using `asyncio.Event()` and `asyncio.wait()` to coordinate the termination of the `MapAsyncIterator` and the underlying async generator.

### DIFF
--- a/graphql/subscription/map_async_iterator.py
+++ b/graphql/subscription/map_async_iterator.py
@@ -44,15 +44,18 @@ class MapAsyncIterator:
             result = await self.iterator.__anext__()
             return self.callback(result)
 
-        _next = ensure_future(self.iterator.__anext__())
         _close = ensure_future(self._close_event.wait())
+        _next = ensure_future(self.iterator.__anext__())
         done, pending = await wait(
-            [_next, _close],
+            [_close, _next],
             return_when=FIRST_COMPLETED,
         )
 
         for task in pending:
             task.cancel()
+
+        if _close.done():
+            raise StopAsyncIteration
 
         if _next.done():
             error = _next.exception()
@@ -63,9 +66,6 @@ class MapAsyncIterator:
                 result = self.reject_callback(error)
             else:
                 result = self.callback(_next.result())
-
-        if _close.done():
-            raise StopAsyncIteration
 
         return (await result) if isawaitable(result) else result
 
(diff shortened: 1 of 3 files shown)


graphql/subscription/map_async_iterator.py
```diff
@@ -19,35 +21,62 @@ def __init__(self, iterable: AsyncIterable, callback: Callable,
...
+            self._close_event.set()
+        else:
+            self._close_event.clear()
 
     def __aiter__(self):
         return self
 
     async def __anext__(self):
-        if self.stop:
+        if self.closed:
```
Cito: Can't we just unconditionally raise StopAsyncIteration here?

graphql/subscription/map_async_iterator.py
```diff
@@ -19,35 +21,62 @@ def __init__(self, iterable: AsyncIterable, callback: Callable,
...
+            error = _next.exception()
+            if error:
+                if not self.reject_callback or isinstance(error, (
+                        StopAsyncIteration, GeneratorExit)):
+                    raise error
+                result = self.reject_callback(error)
+            else:
+                result = self.callback(_next.result())
+
+        if _close.done():
```
Cito: Maybe move this up before the check for _next.done()?

Cito:
Thanks a lot for this contribution. It gets a bit complicated by the usage of Event, but looks ok and I have currently no better idea. Made two small comments in the code above.

graphql/subscription/map_async_iterator.py
dfee: Initially that's what I was doing, but the extra logic was needed to keep this test passing: https://github.com/graphql-python/graphql-core-next/blob/170b227e6cbf67d2cc35555c7ecabbdfcf82d7ac/tests/subscription/test_map_async_iterator.py#L84

Cito: Ah, yes, the finally block should still run.

tests/subscription/test_map_async_iterator.py
```diff
@@ -247,3 +248,27 @@ def __aiter__(self):
...
+        assert result == 2
+
+        # Block at event.wait()
+        fut = ensure_future(anext(doubles))
+        await sleep(.01)
+        assert not fut.done()
+
+        # Trigger cancellation and watch StopAsyncIteration propogate
+        await doubles.aclose()
+        await sleep(.01)
```
Cito: Strange, I need to bump that to .02 to make it pass on my PC.

PR #7, opened 2018-09-06
Added support for middleware

syrusakbary:
Added support for middleware


Cito:
@syrusakbary Can I remove the `middlewares = MiddlewareManager` alias? I guess it had been added for backward compatibility, but in GraphQL-core-next we should get rid of all deprecated stuff.

PR #11, opened 2018-09-21
Await resolvers in parallel

mawa:
This change allows for more parallelism when executing a query. Instead of sequentially `await`ing all the affected resolvers of an object, kick off all the co-routines and then wait for them to finish using `asyncio.gather()`. Effectively, this means you will just wait for the slowest resolver instead of the accumulated time required for all resolvers.

As an illustration, consider the following code which runs twice as fast after this patch is applied:

```python
def async expensive_rpc(*args):
  await asyncio.sleep(1.0)
  return 'hi!'

schema = GraphQLSchema(
  query=GraphQLObjectType(
    name='Root',
    fields={
      'foo': GraphQLField(
        GraphQLString,
        resolve=expensive_rpc,
       ),
       'bar': GraphQLField(
        GraphQLString,
        resolve=expensive_rpc,
       ),
    }
  )
)

await graphql(schema, '{foo, bar'}')
```

### DIFF
--- a/graphql/execution/execute.py
+++ b/graphql/execution/execute.py
@@ -1,3 +1,4 @@
+from asyncio import gather
 from inspect import isawaitable
 from typing import (
     Any,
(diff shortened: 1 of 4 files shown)


mawa:
Oops. Broke the tests, I'll investigate. Please excuse the noise.

Cito:
Thanks, @mawa. You're right, currently the fields are not resolved in parallel, and using `gather()` would solve that. I think you need to pass `return_exceptions=False`, since we want errors to pop up. And maybe you need to fix the order of errors in some tests (which is irrelevant) or fix the test to not check the order (maybe sorting both result and expected). Also, I would just `return dict(await gather(...))` instead of using a dict comprehension.

mawa:
Thanks for your feedback @Cito! I went with your suggestion and sorted the exceptions in the test suite before comparing them. This seems to work okay but the implementation is not perfect (had to convert the relevant properties of `GraphQLError` to a dict).

Cito:
Thanks for the contribution. I'll see if I can simplify the tests a bit.

PR #16, opened 2018-11-19
Keep a list of middleware resolvers

### LINKED ISSUE #15: MiddlewareManager only decorates the first field encountered
`get_middleware_resolvers` is implemented as a generator:

https://github.com/graphql-python/graphql-core-next/blob/f6b078bddae4dd8a48ebf878a6fd5fbbac52bd1b/graphql/execution/middleware.py#L56-L58

The generator object is then assigned to `self._middleware_resolvers`:

https://github.com/graphql-python/graphql-core-next/blob/f6b078bddae4dd8a48ebf878a6fd5fbbac52bd1b/graphql/execution/middleware.py#L30-L32

As the generator is not consumed and unrolled into a list at this point, it is exhausted during the `reduce()` call of the first field construction:

https://github.com/graphql-python/graphql-core-next/blob/f6b078bddae4dd8a48ebf878a6fd5fbbac52bd1b/graphql/execution/middleware.py#L46-L50

As the generator is now exhausted, all other calls to `reduce()` will immediately receive `StopIteration` so no other field is wrapped.

patrys:
Fixes #15 by unrolling the generator into a list.

### DIFF
--- a/graphql/execution/middleware.py
+++ b/graphql/execution/middleware.py
@@ -1,7 +1,7 @@
 from functools import partial, reduce
 from inspect import isfunction
 
-from typing import Callable, Iterator, Dict, Tuple, Any, Optional
+from typing import Callable, Iterator, Dict, List, Tuple, Any, Optional
 
 __all__ = ["MiddlewareManager"]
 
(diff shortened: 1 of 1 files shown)


Cito:
Thanks a lot. We also need to add a regression to graphql-core-next/tests/execution/test_middleware.py that is failing without this fix. If you like, you can add one, otherwise I will do it when I will look into this later this week.

PR #32, opened 2019-05-14
Refactor residual .format()s to f-strings.

changeling:
Replace .format() with f-strings in ast.py and lexer.py.


Cito:
Thank you 👍

PR #35, opened 2019-05-30
Implement __hash__ for GraphQLError

### LINKED ISSUE #34: GraphQLError is unhashable
It seems that the `logging` library in python assumes that exceptions are hashable, in order to be logged.  It'd be great if we could treat GraphQL errors the same way as other builtin exceptions.

See also, a similar issue in the `schematics` project:

https://github.com/schematics/schematics/issues/452

andrew-humu:
Fixes #34 

This seemed like the most minimal implementation that might be safe, but it's definitely possible there's some context I'm missing.  I also claim no mastery of python!

### DIFF
--- a/graphql/error/graphql_error.py
+++ b/graphql/error/graphql_error.py
@@ -166,7 +166,7 @@ class GraphQLError(Exception):
         )
 
     def __hash__(self):
-        return hash(self.original_error)
+        return super(GraphQLError, self).__hash__()
 
     def __ne__(self, other):
         return not self.__eq__(other)
--- a/graphql/error/graphql_error.py
+++ b/graphql/error/graphql_error.py
@@ -79,6 +79,8 @@ class GraphQLError(Exception):
         "extensions",
     )
 
+    __hash__ = Exception.__hash__
+
     def __init__(
         self,
         message: str,
@@ -89,7 +91,7 @@ class GraphQLError(Exception):
         original_error: Exception = None,
         extensions: Dict[str, Any] = None,
     ) -> None:
-        super(GraphQLError, self).__init__(message)
+        super().__init__(message)
         self.message = message
         if nodes and not isinstance(nodes, list):
             nodes = [nodes]  # type: ignore
@@ -164,9 +166,6 @@ class GraphQLError(Exception):
                 for slot in other
             )
         )
-
-    def __hash__(self):
-        return super(GraphQLError, self).__hash__()
 
     def __ne__(self, other):
         return not self.__eq__(other)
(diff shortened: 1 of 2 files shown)


Cito:
Thank your for providing this PR. Hashing only `original_error` assumes that one original error cannot cause two different GraphQLErrors, which is generally true, and that it is always set, which is not always the case. It would be better to derive the hash from all slotted attributes.

But I think the best and most performant solution is to simply use the default hash method for exceptions, based on the id of the instance. GraphQLError lost that default hash capability when it got a custom `__eq__` method, but the default hash method can be restored by setting `__hash__ = Exception.__hash__`. Can you check if that works for you and change the PR accordingly?

andrew-humu:
Your reasoning makes a lot of sense to me.  I pushed a commit — is this what you had in mind?

tests/error/test_graphql_error.py
```diff
@@ -128,9 +128,10 @@ def default_error_formatter_includes_extension_fields():
...
 
     def is_hashable():
-        try:
-            raise RuntimeError("original")
-        except RuntimeError as e:
-            original = e
-        e = GraphQLError("msg", original_error=original)
-        assert hash(e) == hash(original)
+        e = GraphQLError("msg", extensions={"foo": "bar"})
+        assert hash(e) is not Nonw
```
jplock: `None`?

Cito: Python would barf when returning None anyway. Maybe just a one-liner: `hash(GraphQLError("msg"))`

tests/error/test_graphql_error.py
```diff
@@ -128,9 +128,10 @@ def default_error_formatter_includes_extension_fields():
...
+
+    def hashes_are_unique():
```
Cito: Maybe `hashes_are_unique_per_instance` to make this clearer.

tests/error/test_graphql_error.py
```diff
@@ -128,9 +128,10 @@ def default_error_formatter_includes_extension_fields():
...
+        e1 = GraphQLError("msg", extensions={"foo": "bar"})
```
Cito: Maybe remove the extensions arg as it plays no particular role here.

graphql/error/graphql_error.py
```diff
@@ -166,7 +166,7 @@ def __eq__(self, other):
         )
 
     def __hash__(self):
-        return hash(self.original_error)
+        return super(GraphQLError, self).__hash__()
```
Cito: That should be just `super().__hash__()` in Python 3. Or even remove the `def` and simply set  `__hash__ = Exception.__hash__` here.

PR #40, opened 2019-06-29
Set graphql-core-next as v3 of graphql-core

mvanlonden:
This repository will build and distribute the graphql-core package starting at version 3.


graphql/version.py
```diff
@@ -18,9 +18,9 @@ def __str__(self):
         return v
 
 
-version = "1.0.5"
+version = "3.0.0"
```
Cito: There are som more places with the version number, see .bumpversion.cfg

.travis.yml
```diff
@@ -27,3 +27,12 @@ script:
...
 after_success:
    - codecov
+
+deploy:
+  provider: pypi
+  user: syrusakbary
+  on:
+      tags: true
+  password:
+      secure: q7kMxnJQ5LWr8fxVbQPm3pAXKRfYa1d2defM1UXKTQ+Gi6ZQ+QEOAOSbX1SKzYH62+hNRY2JGTeLkTQBeEYn05GJRh+WOkFzIFV1EnsgFbimSb6B83EmM57099GjJnO2nRUU4jyuNGU1joTeaD/g08ede072Es1I7DTuholNbYIq+brL/LQMJycuqZMoWUW4+pP8dE9SmjThMNYHlqNhzdXSE3BlZU0xcw7F2Ea384DNcekIIcapZuPjL167VouuSH/oMQMxBJo+ExEHdbqn5zsA9xcoF931XCgz4ag8U3jHhE48ZXM/xwdQt+S8JnOZcuv3MoAAioMbh+bYXUt2lmENWXCKK1kMDz2bJymwEUeZLA6lFxJQwvlVShowdi7xeyDYLIbeF7yG90Hd+5BqCZn5imzlcQxpjanaQq6xLwAzo6AHssWtd5bBOjDydknPxd1t3QGDoDvtfRdqrfOhlVX5813Hmd/vAopBAba7msKPMLxhsqDZKkwsVrLJLJDjGdpHNl/bbVaMsYcPrsFxa2W8PuddQFviHbL4HDNqHn5SpRwJcQ18YL1X5StQnUz1J+4E0W4mLrU3YW1k8RGlKTes/GeTH4sU+Sh3I9vrDv7849A8U9sSFyB2PT4Jyy8O2R5UyjoqnZDrkYYbLdn/caVo3ThrubTpwdPBmNwcDLA=
```
Cito: As far as I understand, the encrypted passwords are bound to a specific repository. Otherwise anybody could take these and use them in their own repositories. So you will need to re-encrypt the password for the core-next-repository, using the travis CLI client or [travis-encrypt](https://pypi.org/project/travis-encrypt/). If you have admin role on PyPI, you can also give me maintainer role, then I can upload with my account as well.

graphql/version.py
mvanlonden: Ok, I did a find and replace to update the remaining versions.

.travis.yml
mvanlonden: I don't have an admin role on PyPI but have added my encrypted credentials

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.0.0
+current_version = 3.0.0-alpha
```
Cito: Right, we definitely need to publish some alphas and betas before 3.0 final.

But I think it should be 3.0.0a0 (PEP 440).

Cito:
Before making the jump from "Core-Next 1.0.5" to "Core 3.0.0", we should have a working prototype of Graphene 3 in order to be sure we have all necessary functionality (like #41 and #42). I also want to release one last version Core-Next 1.0.6 which prints a deprecation warning and points to Core 3.0.0.

.bumpversion.cfg
mvanlonden: Will make that change

Cito:
Unfortunately, uploading with these deploy credentials failed. Can verify that "travis encrypt" was been executed in the right directory?

PR #44, opened 2019-07-18
Retry pypi password encryption

mvanlonden:
Travis deployment fails with a 403 from PyPI during the deployment step.

Steps taken:
```sh
brew install travis
cd ~/graphql-core-next
travis encrypt <password> --add deploy.password
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,32 @@
 language: python
-
 matrix:
   include:
-    - env: TOXENV=black,flake8,mypy,docs
-      python: 3.7
-      dist: xenial
-      sudo: true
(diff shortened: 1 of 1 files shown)


Cito:
Unfortunately, this still didn't work. Someone [on SO](https://stackoverflow.com/questions/54610805/pypi-deployment-from-travis-ci-invalid-or-non-existent-authentication) advised to add the `--com` option when running travis encrypt. This could actually be the problem, since graphql-core-next was orignially on travis.org and then migrated to travis.com when travis.org became deprecated. So can you make another attemtpt with the `--com` option? If that doesn't help, I'll contact Travis support.

PR #55, opened 2019-09-04
Add pytest-benchmark and port the benchmarks from GraphQL.js

ktosiek:
This is a start for working on #3 - currently only one benchmark is ported from GraphQL.js, but when this code is in an acceptable shape I'll try porting the rest.

Benchmarks can be run with `pytest --benchmark-enable -k benchmark`.
By default benchmarks work like simple tests (they only run once), so they shouldn't slow things down too much.


pyproject.toml
```diff
@@ -45,6 +45,7 @@ sphinx_rtd_theme = ">=0.4"
 check-manifest = ">=0.39"
 bump2version = ">=0.5"
 tox = "^3.13"
+pytest-benchmark = "^3.2"
```
Cito: Can you move this a bit higher up together with the other pytest plugins?

setup.cfg
```diff
@@ -3,3 +3,8 @@ python-tag = py3
 
 [aliases]
 test = pytest
+
+[tool:pytest]
+# Only run benchmarks as tests.
+# To actually run the benchmarks, use --benchmark-enable on the command line.
+addopts = --benchmark-disable
```
Cito: We could also use benchmark-skip to completely skip them, but I agree this is better. They are probably a bit redundant, but not too slow if they run only as test.

tests/utilities/test_introspection_from_schema.py
```diff
@@ -56,3 +61,11 @@ def converts_a_simple_schema_without_description():
             }
             """
         )
+
+    def benchmark_executing_introspection_query(
```
Cito: My suggestion here is to move this to a different file, just like in GraphQL.js.

We could name it `benchmark_introspection_from_schema.py` in the same directory, or move it to a subdirectory `tests/benchmarks/test_introspection_from_schema.py` which I like even better. then you can run the benchmarks with `pytest tests/benchmark --benchmark-enable`, pytest doesn't need to sift through the whole suite to filter the benchmarks, and we have all the benchmarks in one place clearly arranged.

tox.ini
```diff
@@ -41,5 +41,6 @@ deps =
     pytest-asyncio>=0.10
     pytest-cov>=2.7,<3
     pytest-describe>=0.12
+    pytest-benchmark
```
Cito: You should add the same version restriction here as in pyproject.toml

ktosiek:
I've ported the rest of the benchmarks.

Impact on the test suite, as measured with `time`:

| Command | Python 3.6 | Python 3.7|
| ------- | ---------- | --------- |
| `pytest` | 0m11,881s | 0m7,113s |
| `pytest --benchmark-skip` | 0m7,001s | 0m4,311s |
| `pytest --benchmark-enable` | 0m28,314s | 0m17,205s |

So even running the benchmarks as tests adds 3-4s to the whole test suite. Main offenders are:
```
1.13s call     tests/benchmarks/test_introspection_from_schema.py::test_executing_introspection_query
0.67s call     tests/benchmarks/test_validate_gql.py::test_validating_introspection_query
0.64s call     tests/benchmarks/test_validate_sdl.py::test_validating_big_schema_sdl
```

ktosiek:
Update: there was an optimisation in GraphQL.js test setup that I've missed when porting, here are the current times:

| Command | Python 3.6 | Python 3.7|
| ------- | ---------- | --------- |
| `pytest` | 0m10,336s | 0m6,125s |
| `pytest --benchmark-skip` | 0m6,969s | 0m4,252s |
| `pytest --benchmark-enable` | 0m26,654s | 0m16,411s |

So even running the benchmarks as tests adds 2-3s to the whole test suite. Main offenders are:
```
0.69s call     tests/benchmarks/test_introspection_from_schema.py::test_executing_introspection_query
0.65s call     tests/benchmarks/test_validate_sdl.py::test_validating_big_schema_sdl
```

tests/fixtures/__init__.py
```diff
@@ -1,17 +1,22 @@
...
 
 
 def read_graphql(name):
     path = join(dirname(__file__), name + ".graphql")
     return open(path, encoding="utf-8").read()
 
 
+def read_json(name):
+    path = join(dirname(__file__), name + ".json")
+    return json.loads(open(path, encoding="utf-8").read())
```
Cito: You can use json.load instad of json.loads here

tests/fixtures/__init__.py
```diff
@@ -1,17 +1,22 @@
 """Fixtures for graphql tests"""
-
+import json
 from os.path import dirname, join
 
 from pytest import fixture  # type: ignore
 
-__all__ = ["kitchen_sink_query", "kitchen_sink_sdl"]
+__all__ = ["kitchen_sink_query", "kitchen_sink_sdl", "big_schema_sdl"]
```
Cito: The other fixtures should be added here as well

ktosiek:
I've renamed the test functions.
I see there's a conflict in poetry.lock - should I handle that? Are there any poetry-specific tools for handling merges?

Cito:
Thanks. Don't worry, I can resolve that conflict. Will be very busy the next 2 days and look at this later.

Issue #54, opened 2019-09-04
Performance of isawaitable


ktosiek:
I've run a very simplistic benchmark, just returning a long list of single-field objects.
It seems graphql-core-next is 2.5x slower than graphql-core: https://gist.github.com/ktosiek/849e8c7de8852c2df1df5af8ac193287

Looking at flamegraphs, I see isawaitable is used **a lot**, and it's a pretty slow function. Would it be possible to pass raw results around more? It seems resolve_field_value_or_error and complete_value_catching_error are the main offenders here.

Cito:
So far the focus has been on compatibility, correctness, and being up to date with GraphQL.js.

The next step will be to improve performance. There is probably a lot of room for improvement. I'd like to keep the code simple and very close to the GraphQL.js implementation though, in order to make it easier to stay in sync with changes over there, and not move us into a very performant, but inscrutable and unmaintainable dead end.

Maybe we can kind of inline some of the isawaitables() or split some of the code paths for sync/async to avoid such checks.

But before we start the optimization, we need good benchmarks tools. Your script would be a starting point. My idea is to port the existing benchmarks from GraphQL.js and maybe add some more Python specific ones if needed (see issue #3). I think it makes sense t to solve that first, and at least have some basic performance checks before changing the code. 

It's on my list, just need some free time to work on this. Feel free to send contributions as PRs as well.

Cito:
Update: We now have nice benchmarks and can continue working on performance.

qeternity:
> I'd like to keep the code simple and very close to the GraphQL.js implementation though, in order to make it easier to stay in sync with changes over there, and not move us into a very performant, but inscrutable and unmaintainable dead end.

Nice job so far on this project, thanks for all your efforts. I worry that keeping the implementation similar between two very different languages and interpreters is going to keep this project from being scalable. We're using this for a new project, and are considering returning to DRF because for relatively small calls we're struggling to get latencies below 300ms (profiling shows nearly all of the time is spent inside graphql-core). PyPy reduces this by about half, but that means backporting a few repos to 3.6 which I did to test out, but am not keen to maintain especially with 3.8 just around the corner.

Cito:
If it really makes and big difference necessary we certainly can implement the parts that are crucial regarding performance differently. But actually JS and Python are not that different, and as I wrote above, I believe there is still a lot of room for improvement while keeping the implementation similar. Anybody who wants to work on this: Feel free to send PRs and suggestions.

[...]

PR #56, opened 2019-09-05
Use identity instead of hash for caching subfields in collect_subfields

ktosiek:
I've noticed computing `__hash__` on each cache access is quite expensive, and not what the JS implementation does - they are [caching subfield nodes](https://github.com/graphql/graphql-js/blob/e67f2e5515739295bad006ecde0e6ce5db0dc83d/src/execution/execute.js#L1105) based on [argument identity](https://github.com/graphql/graphql-js/blob/e67f2e5515739295bad006ecde0e6ce5db0dc83d/src/jsutils/memoize3.js).

This PR is based on #55 for the benchmarks, but it can work separately.

Results of `pytest --enable-benchmark -k benchmark` before and after the change:
```
------------------------------------------------------ benchmark: 1 tests ------------------------------------------------------
Name (time in s)                               Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------
benchmark_executing_introspection_query     1.1361  1.1782  1.1484  0.0170  1.1427  0.0152       1;1  0.8707       5           1
--------------------------------------------------------------------------------------------------------------------------------
```

```
---------------------------------------------------------- benchmark: 1 tests ----------------------------------------------------------
Name (time in ms)                                Min       Max      Mean  StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------
benchmark_executing_introspection_query     774.3004  789.3146  779.2556  6.0870  778.8992  7.2825       1;0  1.2833       5           1
----------------------------------------------------------------------------------------------------------------------------------------
```

this change also shaved 1/3 of the time from the benchmark mentioned in #54.

### DIFF
--- a/src/graphql/execution/execute.py
+++ b/src/graphql/execution/execute.py
@@ -205,7 +205,7 @@ class ExecutionContext:
         self.middleware_manager = middleware_manager
         self.errors = errors
         self._subfields_cache: Dict[
-            Tuple[GraphQLObjectType, Tuple[FieldNode, ...]], Dict[str, List[FieldNode]]
+            Tuple[GraphQLObjectType, int], Dict[str, List[FieldNode]]
         ] = {}
 
     @classmethod
@@ -998,7 +998,7 @@ class ExecutionContext:
         subfields are not repeatedly calculated, which saves overhead when resolving
         lists of values.
         """
-        cache_key = return_type, tuple(field_nodes)
+        cache_key = return_type, id(field_nodes)
         sub_field_nodes = self._subfields_cache.get(cache_key)
         if sub_field_nodes is None:
             sub_field_nodes = {}


src/graphql/execution/execute.py
```diff
@@ -998,7 +996,7 @@ def collect_subfields(
         subfields are not repeatedly calculated, which saves overhead when resolving
         lists of values.
         """
-        cache_key = return_type, tuple(field_nodes)
+        cache_key = id(return_type), id(field_nodes)
```
Cito: The first element of the tuple actually does not need id() to be called, since it uses object identity anyway. It's only needed for the list which is not hashable resp. too slow to hash after conversion to a tuple. Can you change that and adapt the type hint above accordingly?

Cito:
👍 Thank you.

Issue #61, opened 2019-09-15
SourceLocation should be serialized as object


ktosiek:
`SourceLocation`, as a NamedTuple, is serialized by json.dumps() to an array instead of an object. Graphql.js [keeps `SourceLocation` as and object](https://github.com/graphql/graphql-js/blob/57232347e71b7b23fbcd4c53c8afca73f89e9cc7/src/language/location.js#L8), so it just works for them.

Reproduction:
```python
from graphql import SourceLocation, GraphQLError, format_error, Source
import json

print(json.dumps(format_error(GraphQLError(message="test", source=Source('{ test }'), positions=[2]))))
```

Expected result: `{"message": "test", "locations": [`**`{"line": 1, "column": 3}`**`], "path": null}`
Actual result: `{"message": "test", "locations": [`**`[1, 3]`**`], "path": null}`

Should `SourceLocation` be changed to something that serializes correctly? Or would it make more sense to convert it to dict in `format_errors`?

Cito:
Good point. Yes, I think it makes totally sense to return locations as a list of dicts in `format_error` - that's why we have this function. Feel free to send a PR, or otherwise I can change this.

Cito:
This is now included in [v3.0.0b1](https://github.com/graphql-python/graphql-core-next/releases/tag/v3.0.0b1).

PR #62, opened 2019-09-16
format_error should return locations as a list of {"line", "column"}

ktosiek:
Fix for #61. This way a simple json.dumps() of format_error(...) should generate correct locations.

I've left `GraphQLError.__eq__` as is, but maybe it should compare dicts to self.formatted?

### DIFF
--- a/src/graphql/error/format_error.py
+++ b/src/graphql/error/format_error.py
@@ -17,7 +17,11 @@ def format_error(error: "GraphQLError") -> Dict[str, Any]:
         raise TypeError("Received no error object.")
     formatted: Dict[str, Any] = dict(  # noqa: E701 (pycqa/flake8#394)
         message=error.message or "An unknown error occurred.",
-        locations=error.locations,
+        locations=(
+            [{"line": l.line, "column": l.column} for l in error.locations]
+            if error.locations is not None
+            else None
+        ),
         path=error.path,
     )
     if error.extensions:
--- a/src/graphql/error/format_error.py
+++ b/src/graphql/error/format_error.py
@@ -18,7 +18,7 @@ def format_error(error: "GraphQLError") -> Dict[str, Any]:
     formatted: Dict[str, Any] = dict(  # noqa: E701 (pycqa/flake8#394)
         message=error.message or "An unknown error occurred.",
         locations=(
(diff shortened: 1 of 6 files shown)


tests/error/test_format_error.py
```diff
@@ -30,9 +30,10 @@ def format_graphql_error():
             ValueError("original"),
             extensions=extensions,
         )
-        assert error == {
+        formatted = format_error(error)
```
Cito: Good catch!

src/graphql/error/format_error.py
```diff
@@ -17,7 +17,11 @@ def format_error(error: "GraphQLError") -> Dict[str, Any]:
         raise TypeError("Received no error object.")
     formatted: Dict[str, Any] = dict(  # noqa: E701 (pycqa/flake8#394)
         message=error.message or "An unknown error occurred.",
-        locations=error.locations,
+        locations=(
+            [{"line": l.line, "column": l.column} for l in error.locations]
```
Cito: Better use `loc` or `location` instead of the letter `l` - better readable and cannot be confused with the digit 1.

To be consistent in style with the outer dict, better write it like `dict(line=loc.line, column=loc.column)`, or rewrite the outer dict as literal expression.

We could also set `locations=[loc,_asdict() for loc in error.locations]`, but that will produce an OrderedDict, while we use ordinary dicts everywhere else (since they are ordered in Py 3.6+ anyway).

ktosiek:
codecov doesn't like the `SourceLocation.__ne__` method being untested. Should I add explicit tests for `__eq__` and `__ne__`?

src/graphql/language/location.py
```diff
@@ -12,6 +12,20 @@ class SourceLocation(NamedTuple):
...
+    @property
+    def formatted(self):
+        return dict(line=self.line, column=self.column)
+
+    def __eq__(self, other):
+        if isinstance(other, dict):
+            return other == self.formatted
+        return super().__eq__(other)
+
+    def __ne__(self, other):
```
Cito: Good catch to add `__ne__`. We need that to override the builtin method. But I think `return not self.__eq__(other)` should suffice here.

Cito:
Thank you!

PR #66, opened 2019-11-22
Add test for problem with coerce_input_value

ganwell:
While working on graphql-python/graphene-django#812 I found that coerce_input_value will pollute the input value with every field of the type set to None. This will then fail in filtering.

I don't know where to fix that correctly. I "fixed" it like that:

```
a/src/graphql/utilities/coerce_input_value.py
```
```python
@@ -90,7 +90,7 @@ def coerce_input_value(
             field_value = input_value.get(field_name, INVALID)

             if field_value is INVALID:
-                if field.default_value is not INVALID:
+                if field.default_value not in (INVALID, None):
                     # Use out name as name if it exists (extension of GraphQL.js).
                     coerced_dict[field.out_name or field_name] = field.default_value
                 elif is_non_null_type(field.type):
```
Which is not correct, since the documentation says that it should be coerced to None. So the problem is actually a mismatch between expectations of the filter and how coerce_input_value is defined.

I appreciate any pointers or hints.

### DIFF
--- a/tests/execution/test_variables.py
+++ b/tests/execution/test_variables.py
@@ -176,6 +176,22 @@ def describe_execute_handles_inputs():
                     None,
                 )
 
+            def properly_coerces_input_varaibles():
+                result = execute_query(
+                    """
+                    query FieldQuery($input: TestInputObject){
+                      fieldWithObjectInput(input: $input)
+                    }
+                    """,
+                    variable_values={"input": {'a': "foo", 'b': "bar", 'c': "baz"}}
+                )
+                assert result == (
+                    {
+                        'fieldWithObjectInput': "{'a': 'foo', 'b': ['bar'], 'c': 'baz'}"
+                    },
+                    None
+                )
+
             def properly_parses_null_value_to_null():
                 result = execute_query(
                     """


Cito:
Hi @ganwell. Thanks for contributing! Currently, I'm not working with graphene-django and have no time to dig into that, so you need to give me some more clues. What exactly do you mean with "coerce_input_value will pollute the input value with every field of the type set to None"? If you don't pass a type, but a non-null value then you should get an error. Can you give an example? With which arguments do you call coerce_input_value? What do you expect, what do you get instead? The test that you just added passes for me without changing anything. And the line that you changed in coerce_input_value looks correct to me, GraphQL.js also does not check for null (None) here. Note that INVALID is used as the equivalent of "undefined" in GraphQL.js. Setting a default value to None is different from the "default default value" INVALID which means "there is no default value specified".

ganwell:
@Cito Thanks for answering so quickly. In my original issue graphene was involved. I created the unittest in this PR to mimic that issue. I don't know how I was able to get failures with graphql-core only. Maybe some debug code that I forgot to reset. Graphene injects bad (None) default_values, I am investigating in graphene now. My goal is to update our projects that use graphane-django to a 3.0 stack.

Cito:
@ganwell I think Graphene/Core 2 did not support the distinction between null and undefined default values. So you probably need to make some adaptations for Core 3, either in Graphene 3 or in graphene-django. If you want NO default value in Core 3, then pass nothing or INVALID, but do not pass None as default value.

PR #74, opened 2020-01-05
Use asyncio.iscoroutine instead of inspect.isawaitable

corydolphin:
R.e. the performance issues seen here: https://github.com/graphql-python/graphql-core-next/issues/54

Long term, it seems the best fix would be to minimize the usage of `iscoroutine`, in favor of inspecting and caching whether or not the function `iscoroutinefunction`. 

As a shorter term fix, this change switches form `inspect.isawaitable` to `asyncio.iscoroutine`. `inspect.isawaitable` is pure Python, while asyncio.iscoroutine is optimized in C and has a smarter type-cache. Reading through their implementations and comments, it isn't clear to me whether or not there is a semantic difference between the two, I'm relying on integration tests here to catch that. If there is a difference, I can update to use an early-exiting `or` to at least take the fast path when possible, assuming standard coroutines are the common case 🤞

### DIFF
--- a/src/graphql/execution/execute.py
+++ b/src/graphql/execution/execute.py
@@ -1,5 +1,4 @@
-from asyncio import gather
-from inspect import isawaitable
+from asyncio import gather, iscoroutine
 from typing import (
     Any,
     Awaitable,
@@ -303,7 +302,7 @@ class ExecutionContext:
         Given a completed execution context and data, build the (data, errors) response
         defined by the "Response" section of the GraphQL spec.
         """
-        if isawaitable(data):
+        if iscoroutine(data):
 
             async def build_response_async():
                 return self.build_response(await data)  # type: ignore
@@ -351,7 +350,7 @@ class ExecutionContext:
             self.errors.append(error)
             return None
         else:
-            if isawaitable(result):
+            if iscoroutine(result):
                 # noinspection PyShadowingNames
                 async def await_result():
                     try:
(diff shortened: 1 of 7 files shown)


corydolphin:
Darn. The one issue is that `iscoroutine` returns `True` if the value is a generator, this breaks the list resolver in the case of a generator.

Cito:
Thanks for your PR. I'm currently very busy, but will look into this when I find time. Maybe we can find a solution for that issue with generators.

syrusakbary:
Coroutines and awaitables are different things.

Coroutines refer to async functions, while awaitables refer to the result of running async functions.

isawaitable is the right thing to ask for, if you are asking in the data. We could ask on `iscoroutine` in the function, but this make the following to not work:

```python
async def my_async_func():
  return 1

def resolve_x(root, info, **args):
  return my_async_func()

assert iscoroutine(resolve_x) == False
assert isawaitable(resove_x()) == True
```

Closing the PR

Cito:
As was already noticed, the difference is that `isawaitable` also checks for generator-based coroutines and the existence of the `__await__` method which coroutines do not have. But the approach to avoid the wasteful isinstance(Awaitable) check should still be investigated. This is discussed in #54.

Cito:
@syrusakbary `asyncio.iscoroutine` actually checks for coroutine *objects*. In your example, you do not call `resolve_x()` in the first ` assert` statement, otherwise it would be equal to True.

I have removed my last misleading comment. The problem with `asyncio.iscoroutine` is *not*  that it does not detect old-syle generator-based coroutine.

But the problems with `asyncio.iscoroutine` instead of `isawaitable` *are*:
1) it returns True for generators (as @corydolphin already commented),
2) it returns False for Futures because they are not coroutines.

PR #82, opened 2020-03-17
Make nodes extensible and weak-referencable

### LINKED ISSUE #81: Weakref unsupported in AST
`graphql.language.ast` makes extensive use of `__slots__` but `__weakref__` is given, breaking the `weakref` module.

* [Python Documentation about `__slots__`](https://docs.python.org/3/reference/datamodel.html#slots)

The use of `__slots__` without `__weakref__` makes it very hard to track data about AST nodes in a non-leaky way.

astronouth7303:
Add standard dunder attributes to (almost) all instances of `__slots__`

Using the test suite as a benchmark, this change produced no noticeable slow down.

Closes #81

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,11 +191,10 @@ Design goals for the GraphQL-core 3 library are:
 
 Some restrictions (mostly in line with the design goals):
 
-* requires Python 3.6 or 3.7
+* requires Python 3.6 or newer
 * does not support some already deprecated methods and options of GraphQL.js
 * supports asynchronous operations only via async.io
   (does not support the additional executors in GraphQL-core)
-* the benchmarks have not yet been ported to Python
 
 
 ## Integration with other libraries and roadmap
--- a/src/graphql/execution/middleware.py
+++ b/src/graphql/execution/middleware.py
@@ -20,7 +20,7 @@ class MiddlewareManager:
     must be aware of this and check whether values are awaitable before awaiting them.
     """
 
-    __slots__ = "middlewares", "_middleware_resolvers", "_cached_resolvers"
+    __slots__ = "__weakref__", "__dict__", "middlewares", "_middleware_resolvers", "_cached_resolvers"
 
     _cached_resolvers: Dict[GraphQLFieldResolver, GraphQLFieldResolver]
     _middleware_resolvers: Optional[List[Callable]]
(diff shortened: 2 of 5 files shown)


astronouth7303:
I did not touch `src/graphql/validation/rules/overlapping_fields_can_be_merged.py`

Cito:
> I did not touch `src/graphql/validation/rules/overlapping_fields_can_be_merged.py`

Good, that PairSet class is only for internal usage anyway.

Cito:
Thanks for sending a PR. Not quite sure if we really should add these to all classes, like `Location` (attached to every node) and `Token` (mainly for internal use in the lexer) extendible.
Would like to get @syrusakbary's opinion regarding this PR.

Cito:
In order to get a better feeling for the memory penalty, I ran memory_profiler while parsing the big_schema_sdl now, and found that for each of the 3 classes in the ast module, it uses 4% more memory, 12% more in total. I think this is pretty significant already. So I suggest relaxing only the Node class where it makes most sense. What do you think?

astronouth7303:
Sure, I can do that. My use case is primarily about `Node` anyway.

astronouth7303:
How do I do the memory profiling?

Cito:
Load `github_schema`, run `parse()` on it, look at the memory increase in that line with [memory_profiler](https://github.com/pythonprofilers/memory_profiler).

astronouth7303:
Ok, some variations. Using [this script](https://gist.github.com/astronouth7303/16e93e0c692caaa226a47d8af65e4673), each is an average of 3 sequential runs

* `master`: 42.2MB / 17.8MB
* This PR, as-is: 44.6MB / 19.9MB
* Variation 1: 43.1MB / 18.5MB
* Variation 2: 43.9 MB / 19.3MB
* Variation 3: 43.9MB / 19.3MB
* Variation 4: 44.3MB / 19.8MB

Ok, variations:
* Variation 1: Dropped changes to `Location` and `Token`
* Variation 2: Dropped changes to `Token`
* Variation 3: Dropped `__dict__` on `Location` and `Token`
* Variation 4: Dropped `__dict__` on `Token`

I was surprised by the memory impact, even between master and variation 3 (presence or removal of `__weakref__`). All that memory is _just_ space from the extra slots--`__weakref__` is initialized to `None`, which is a singleton. (Under the hood in CPython, it's an actual value, `PyNone`; `PyObject*` is not normally `NULL`.)

I guess I'll go with variation 1, then.

Cito:
Thanks a lot for the detailed verification. I agree variation 1 is the best compromize.

astronouth7303:
Ok, I think that's all squared away.

PR #87, opened 2020-04-11
Fix single backticks in docs

hoefling:
If a crossreference is possible, single backticks are replaced with the suitable crossref. Otherwise, single backticks are replaced with double backticks. I have tried to keep the diff as small as possible, avoiding rewrapping of long lines.


Cito:
Thank you @hoefling for working on this. The improved Sphinx documentation probably outweighs the decreased readability of the raw docstrings.

PR #92, opened 2020-05-15
Fix method name in Visitor docstring

nawatts:
The docstring for the Visitor class describes the signatures for the `enter` and `leave` methods. However, it labels both of them as `enter`.


Cito:
Oops, thank you!

PR #103, opened 2020-08-15
coerce str

ryanwang520:
It's convenient for frontend to pass `str` as some  value of `Int`  type,  maybe it's too strict to reject valid  **int str** .

By the way, I tried to upgrade a project from `graphene 2` to `graphene 3` and found that in `graphene 2`  str would coerce to int correctly while `graphene 3` just throws the error `Int cannot represent non-integer value: `, which I think is caused by the `coerce_int` function here.

### DIFF
--- a/src/graphql/type/scalars.py
+++ b/src/graphql/type/scalars.py
@@ -1,5 +1,5 @@
 from math import isfinite
-from typing import Any
+from typing import Any, Union
 
 from ..error import GraphQLError
 from ..pyutils import inspect, is_finite, is_integer, FrozenDict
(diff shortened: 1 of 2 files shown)


Cito:
Thanks for contributing.

I looked into this and found that the addressed behavior and tests in GraphQL-core are the [same as in GraphQL.js](https://github.com/graphql/graphql-js/blob/7aab39c6a9df887136de8c3f7d0e9c9995b7319f/src/type/__tests__/scalars-test.js#L50). Since the goal of GraphQL-core is to be a faithful replication of GraphQL.js which is considered the "reference implementation", I really do not want to introduce incompatible changes in GraphQL-core. If you think this really needs to be changed, you have two options: 1. Create an issue in the GraphQL.js project. and convince them to implement the change there  - don't worry, they are open for such suggestions and discussions. If it is accepted, it will then be ported to GraphQL-core as well. 2. Try to change it at a higher level, e.g. in Graphene.

ryanwang520:
Ok, I would just try some other workaround.  Thanks.

Issue #99, opened 2020-07-01
Make use of TypedDict


Cito:
We could make use of [TypedDict](https://www.python.org/dev/peps/pep-0589/) in several places, for instance as return value of  `to_kwargs` and `introspection_from_schema`.

Unfortunately, this needs Python 3.8, but we are currently supporting Python 3.6. However, it should be possible to add `typing_extensions` as additional dependency for Python 3.6 and 3.7 to support TypedDict for these Python versions as well.

KingDarBoja:
Started to use `TypedDict` on **graphql-server-core**, those are awesome to work with and the `typing_extensions` does the job to bring backward compatibility to Python 3.6 and 3.7. Only thing that you must be aware is that you must provide all values to any TypeDict declaration as those doesn't support default values on class definition as far as I know.

Cito:
This has now been implemented for `introspection_from_schema` in 24506f31ed1dbd104d0f5481965d0e0814f127f1.

Still needs to be implemented for `to_kwargs`.

Cito:
This is now also implemented for `to_kwargs`, so we can close this issue.

PR #108, opened 2020-09-19
refactor: WIP introspection from schema using typeddict

KingDarBoja:
Work in progress based on #99 request.

cc @Cito

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.6"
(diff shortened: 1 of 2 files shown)


Cito:
Thank you for the contribution, finally came back to this and implemented it in 24506f31ed1dbd104d0f5481965d0e0814f127f1 using TypedDicts, since we are using dicts as introspection results. Unfortunately, the implementation of TypedDicts in mypy has some weaknesses (no recursion, no generics) so this is not as good as it could be. Closing this now.

Issue #116, opened 2020-11-09
Performance of validate, in the context of graphql(_sync)


alexchamberlain:
I've benchmarked a few of our use cases and up to 85% of the runtime is in `validation.validate`; as far as I can tell, there's been no discussion of this on Issues, so I thought I'd open an issue to see if this is a known issue?

I'll follow up with a PR that demonstrates the issue and has some improvements, but won't be production quality at this stage.

Cito:
Hi @alexchamberlain. Not sure if we have explicitly discussed this already. One issue is always when and how to avoid the validation step and not validating the same query multiple times (#70 is related to that). 

I'm glad about PRs that try to improve performance. However, the code should not deviate too much from the original code in GraphQL.js since this will make it difficult to merge future change in GraphQL.js back here. If you want to introduce larger algorithmic changes, it may make more sense to suggest them on the GraphQL.js issue tracker, implement these in JavaScript and then merge back here.

alexchamberlain:
Thanks @Cito 

I've opened #117 with the benchmark script I used. The main change is to memoise the lookup of visit functions, which is more of a Python optimisation I think?

Cito:
Thanks. I will look into this if I find some time. Please note that there are already some benchmark scripts. It would make sense to add another benchmark that explicitly focuses on validation then.

alexchamberlain:
Ah, sorry - I completely missed that! I can look to refactor my benchmarks there if you like?

Cito:
Sorry for leaving this open for so long. I think it has now been addressed and solved in ac3caf60fe9f2dd2ef70bcd530ed0e0d0c54b638 and #150.

Issue #70, opened 2019-12-06
Allow passing rules directly into the graphql function


asgoel:
I wanted to add some custom validation rules in addition to the default `graphql.specified_rules`. Unless I'm mistaken, the only way to do so is to manually call `validate` with the new rules, which means that I basically have to duplicate all the code inside of `async def graphql`, and change the validation step in order to pass in my custom rules. Is it possible instead to just allow clients to pass the rules they want to use into the top-level `graphql` call?

Cito:
@asgoel, thanks for reporting the issue.

As far as I see, you really need to build your own graphql() function to do that.

But I have forwarded your question to the issue tracker of GraphQL.js - maybe they have an idea or will add such an argument to the graphql() function (graphql/graphql-js#2278) - please join the discussion there as well. The preferred way to get new features into graphql-core is to get them into the reference implementation GraphQL.js, of which graphql-core is a port. That way everyone will benefit and graphql-core stays compatible.

PR #117, opened 2020-11-09
validation.validate slowness

alexchamberlain:
Small scratch script demonstrating validation.validate slowness.

You can run by installing graphql-core in a venv, then running scratch.py. Personally,
I then run it through gprof2dot for visualisation

Demonstration of #116 

By memoising the various visitors, I got the runtime proportion down from ~85% to ~70%. Removing the variable checks removed another 10%. I'm not sure how safe either of these were.

### DIFF
--- a/src/graphql/language/visitor.py
+++ b/src/graphql/language/visitor.py
@@ -173,6 +173,9 @@ class Visitor:
     # Provide special return values as attributes
     BREAK, SKIP, REMOVE, IDLE = BREAK, SKIP, REMOVE, IDLE
 
+    def __init__(self):
+        self._visit_fns = {}
+
     def __init_subclass__(cls) -> None:
         """Verify that all defined handlers are valid."""
         super().__init_subclass__()
@@ -197,11 +200,12 @@ class Visitor:
 
     def get_visit_fn(self, kind: str, is_leaving: bool = False) -> Callable:
         """Get the visit function for the given node kind and direction."""
-        method = "leave" if is_leaving else "enter"
-        visit_fn = getattr(self, f"{method}_{kind}", None)
-        if not visit_fn:
-            visit_fn = getattr(self, method, None)
-        return visit_fn
+        key = (kind, is_leaving)
+        if key not in self._visit_fns:
+            method = "leave" if is_leaving else "enter"
+            fn = getattr(self, f"{method}_{kind}", None)
+            self._visit_fns[key] = fn or getattr(self, method, None)
+        return self._visit_fns[key]
 
 
 class Stack(NamedTuple):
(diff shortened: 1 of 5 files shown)


alexchamberlain:
Correction: Obviously, the variable checks removal is not safe. I just did that to inspect how expensive the recursive visitor was.

src/graphql/utilities/type_info.py
```diff
@@ -136,15 +137,25 @@ def get_enum_value(self) -> Optional[GraphQLEnumValue]:
...
             method()
 
+    def _get_method(self, direction: str, kind: str) -> Optional[Callable[[], None]]:
+        key = (direction, kind)
+        if key in self._visit_fns:
+            return self._visit_fns[key]
+
+        fn = getattr(self, f"{direction}_{kind}", None)
+        self._visit_fns[key] = fn
+        return fn
```
syrusakbary: I think there is a cleaner way to do this:

```suggestion
        if key not in self._visit_fns:
            fn = getattr(self, f"{direction}_{kind}", None)
            self._visit_fns[key] = fn
        return self._visit_fns[key]
```

src/graphql/validation/specified_rules.py
```diff
@@ -118,15 +118,15 @@
         PossibleFragmentSpreadsRule,
         NoFragmentCyclesRule,
         UniqueVariableNamesRule,
-        NoUndefinedVariablesRule,
-        NoUnusedVariablesRule,
+        # NoUndefinedVariablesRule,
+        # NoUnusedVariablesRule,
```
syrusakbary: This should be reverted

src/graphql/validation/specified_rules.py
```diff
@@ -118,15 +118,15 @@
...
         KnownDirectivesRule,
         UniqueDirectivesPerLocationRule,
         KnownArgumentNamesRule,
         UniqueArgumentNamesRule,
         ValuesOfCorrectTypeRule,
         ProvidedRequiredArgumentsRule,
-        VariablesInAllowedPositionRule,
+        # VariablesInAllowedPositionRule,
```
syrusakbary: This should be reverted

src/graphql/language/visitor.py
```diff
@@ -197,10 +200,18 @@ def __init_subclass__(cls) -> None:
...
+            return self._visit_fns[key]
+
         method = "leave" if is_leaving else "enter"
         visit_fn = getattr(self, f"{method}_{kind}", None)
         if not visit_fn:
             visit_fn = getattr(self, method, None)
+
+        self._visit_fns[key] = visit_fn
+
         return visit_fn
```
syrusakbary: This can be done more cleanly:

```suggestion
        key = (kind, is_leaving)
        if key not in self._visit_fns:
            method = "leave" if is_leaving else "enter"
            fn = getattr(self, f"{method}_{kind}", None)
            self._visit_fns[key] = fn or getattr(self, method, None)
        return self._visit_fns[key]
```

src/graphql/validation/specified_rules.py
Cito: Yes, the default validation behavior should be the same as in GraphQL.js which reflects the official specs. If you want to omit certain validation rules, see #70.

src/graphql/language/visitor.py
Cito: LGTM with the change proposed by Syrus.

A `try ... except KeyError` might be even faster depending how often the visitor is reused. You may want to do some benchmarking here.

src/graphql/validation/specified_rules.py
Cito: See below, the default rules should not be changed.

src/graphql/utilities/type_info.py
Cito: LGTM, see remark regarding visitor.py

src/graphql/validation/specified_rules.py
alexchamberlain: Sorry, this was more to demonstrate that these rules are quite expensive compared to the others.

I wonder if these could be separated into "essential" and "friendly". For example, I think `NoUnusedVariablesRule` helps a user develop a query, but may not be needed in a production environment where there isn't a developer to provide feedback to.

pyproject.toml
```diff
@@ -52,5 +52,5 @@ tox = "^3.19"
 target-version = ['py36', 'py37', 'py38']
 
 [build-system]
-requires = ["poetry>=1,<2"]
+requires = ["poetry>=1,<2", "setuptools"]
```
alexchamberlain: This was added to make `pip install -e .` work; I need to research whether `poetry` has something built in. Not really for landing.

Cito:
This has been addressed in ac3caf60fe9f2dd2ef70bcd530ed0e0d0c54b638 and #150 now.

PR #127, opened 2021-04-09
This fixes embedded async resolvers for Ariadne 

ammurdoch:
I ran into an issue while using Ariadne where embedded async resolvers were not getting awaited and causing my queries to fail.  Top level async resolvers worked fine, but resolvers two or three levels down weren't getting awaited.  I dug into the issue and was able to fix it by adding the following to the `resolve_field_value_or_error` function:

```diff
def resolve_field_value_or_error(
        self,
        field_def: GraphQLField,
        field_nodes: List[FieldNode],
        resolve_fn: GraphQLFieldResolver,
        source: Any,
        info: GraphQLResolveInfo,
    ) -> Union[Exception, Any]:
        """Resolve field to a value or an error.

        Isolates the "ReturnOrAbrupt" behavior to not de-opt the resolve_field()
        method. Returns the result of resolveFn or the abrupt-return Error object.

        For internal use only.
        """
        try:
            # Build a dictionary of arguments from the field.arguments AST, using the
            # variables scope to fulfill any variable references.
            args = get_argument_values(field_def, field_nodes[0], self.variable_values)

            # Note that contrary to the JavaScript implementation, we pass the context
            # value as part of the resolve info.
            result = resolve_fn(source, info, **args)
            if self.is_awaitable(result):
                # noinspection PyShadowingNames
                async def await_result() -> Any:
                    try:
-                        return await result
+                        intermediate_result = await result
+                        if isawaitable(intermediate_result):
+                             return await intermediate_result
+                        return intermediate_result
                    except Exception as error:
                        return error

                return await_result()
            return result
        except Exception as error:
            return error
```

### DIFF
--- a/src/graphql/execution/execute.py
+++ b/src/graphql/execution/execute.py
@@ -639,7 +639,10 @@ class ExecutionContext:
                 # noinspection PyShadowingNames
                 async def await_result() -> Any:
                     try:
-                        return await result
+                        intermediate_result = await result
+                        if isawaitable(intermediate_result):
+                            return await intermediate_result
+                        return intermediate_result
                     except Exception as error:
                         return error
 


Cito (commented):
* Can you provide an executable example (that doesn't involved Ariadne)? You don't need to make an actual unit test from this, I can do this.
* It looks like this works only for two levels of nesting - if we do something like this, it should work for deeper nesting as well.
* The actual problem may be that an await is missing in the nested resolvers. Therefore we really need some example code.

Cito:
Closing this since there was no more feedback. If you still think there is a problem, please open a new issue with a reproducable example.

PR #130, opened 2021-04-28
Updated ExecutionResult.formatted

### LINKED ISSUE #129: ExecutionResult.formatted does not format errors
When I call formatted on an ExecutionResult, the errors portion of the dict is not formatted and instead returns a list of GraphQLErrors. GraphQLError is not serializable, but it does have a formatted property which is, why is formatted not called on the list of GraphQLErrors?

As a work around, I have to call formatted on my ExecutionResult and then rebuild the dict and format the errors:
```
    formatted = result.formatted

    data = formatted.get('data')
    errors = ([err.formatted for err in formatted.get('errors')] 
              if formatted.get('errors') is not None 
              else None)
    extensions = formatted.get('extensions')

    if extensions:
        body = dict(data=data, errors=errors, extensions=extensions)
    else:
        body = dict(data=data, errors=errors)
```

Ideally ExecutionResult.formatted would perform this error formatting for me.

Environment:
python 3.8
graphql-core 3.1.4

mlorenzana:
Updated formatted property to chain formatted of errors

Fixes #129

### DIFF
--- a/src/graphql/execution/execute.py
+++ b/src/graphql/execution/execute.py
@@ -131,9 +131,12 @@ class ExecutionResult:
     @property
     def formatted(self) -> Dict[str, Any]:
         """Get execution result formatted according to the specification."""
+        formatted_errors = ([err.formatted for err in self.errors]
+                            if self.errors is not None
+                            else None)
         if self.extensions is None:
-            return dict(data=self.data, errors=self.errors)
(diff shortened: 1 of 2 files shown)


Cito:
That's pretty much how I would have done it as well. Thanks for the contribution.

PR #138, opened 2021-09-02
Set enum value values to value names in build_client_schema

charmasaur:
This makes it consistent with build_ast_schema (see commit
2a1953f8563135b85bd7b0ecf78e7630484ccdb2), and ensures that default enum
values are reflected in schemas built from introspection when using
`gql` (the introspection version of
https://github.com/graphql-python/graphql-core/issues/111).

### DIFF
--- a/src/graphql/utilities/build_client_schema.py
+++ b/src/graphql/utilities/build_client_schema.py
@@ -175,6 +175,7 @@ def build_client_schema(
             description=enum_introspection.get("description"),
             values={
                 value_introspect["name"]: GraphQLEnumValue(
+                    value=value_introspect.get("name"),
                     description=value_introspect.get("description"),
                     deprecation_reason=value_introspect.get("deprecationReason"),
                 )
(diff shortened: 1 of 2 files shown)


src/graphql/utilities/build_client_schema.py
```diff
@@ -175,6 +175,7 @@ def build_enum_def(enum_introspection: Dict) -> GraphQLEnumType:
             description=enum_introspection.get("description"),
             values={
                 value_introspect["name"]: GraphQLEnumValue(
+                    value=value_introspect.get("name"),
```
Cito: You can change this to value_introspect["name"] since the name must be always set and we do this for the key as well.

Cito (changes requested):
Thank you for the contribution - I agree this is the right thing to do.

Will probably also backport this to the 3.1 branch.

Cito (approved):

src/graphql/utilities/build_client_schema.py
charmasaur: D'oh, good point, done!

charmasaur:
Thanks for the quick review+merge!

PR #145, opened 2021-10-18
Create async loop outside of benchmark

jkimbo:
I realised that the async benchmark in #141 wasn't a fair comparison to the sync one because it was also including the async loop setup. This PR creates the async loop outside of the benchmark code to more accurately benchmark the actual query execution.

There is still some overhead with the async execution but it's no longer as bad as the first benchmark (more like ~20% slower rather than ~70%)
CleanShot 2021-10-18 at 11 57 22@2x


Cito (changes requested):
That's a really useful observation and fix.

Can you add a small comment here in the code as well, explaining why it is important that you create the loop outside?

Cito:
@jkimbo Does this also apply to the SQLAlchemy benchmark?

jkimbo:
> @jkimbo Does this also apply to the SQLAlchemy benchmark?

@Cito I don't think so because an async server would have created the event loop on start up so you wouldn't be creating a new one per request.

PR #146, opened 2021-11-05
chore: Refactor MultiCode Expression ✨

yezz123:
I just refactor some functions, mostly was simple like:
- Swap if & else branches.
- Merge Else If Into Elif.
- De Morgan Identity.

### DIFF
--- a/src/graphql/language/block_string.py
+++ b/src/graphql/language/block_string.py
@@ -38,7 +38,7 @@ def dedent_block_string_value(raw_string: str) -> str:
 
 def is_blank(s: str) -> bool:
     """Check whether string contains only space or tab characters."""
-    return all(c == " " or c == "\t" for c in s)
+    return all(c in [" ", "\t"] for c in s)
 
 
 def get_block_string_indentation(value: str) -> int:
--- a/src/graphql/language/block_string.py
+++ b/src/graphql/language/block_string.py
@@ -38,7 +38,7 @@ def dedent_block_string_value(raw_string: str) -> str:
 
 def is_blank(s: str) -> bool:
     """Check whether string contains only space or tab characters."""
-    return all(c in [" ", "\t"] for c in s)
+    return all(c == " " or c == "\t" for c in s)
 
 
 def get_block_string_indentation(value: str) -> int:
--- a/src/graphql/language/visitor.py
+++ b/src/graphql/language/visitor.py
@@ -183,16 +183,15 @@ class Visitor:
                 kind: Optional[str] = None
             else:
                 attr, kind = attr_kind
-            if attr in ("enter", "leave"):
-                if kind:
-                    name = snake_to_camel(kind) + "Node"
-                    node_cls = getattr(ast, name, None)
-                    if (
-                        not node_cls
-                        or not isinstance(node_cls, type)
-                        or not issubclass(node_cls, Node)
-                    ):
-                        raise TypeError(f"Invalid AST node kind: {kind}.")
+            if attr in ("enter", "leave") and kind:
+                name = snake_to_camel(kind) + "Node"
+                node_cls = getattr(ast, name, None)
+                if (
+                    not node_cls
+                    or not isinstance(node_cls, type)
+                    or not issubclass(node_cls, Node)
+                ):
+                    raise TypeError(f"Invalid AST node kind: {kind}.")
 
     def get_visit_fn(self, kind: str, is_leaving: bool = False) -> Callable:
         """Get the visit function for the given node kind and direction."""
@@ -256,10 +255,7 @@ def visit(root: Node, visitor: Visitor) -> Any:
             node: Any = parent
             parent = ancestors_pop() if ancestors else None
             if is_edited:
-                if in_array:
-                    node = node[:]
-                else:
-                    node = copy(node)
+                node = node[:] if in_array else copy(node)
             edit_offset = 0
             for edit_key, edit_value in edits:
                 if in_array:
@@ -267,11 +263,10 @@ def visit(root: Node, visitor: Visitor) -> Any:
                 if in_array and (edit_value is REMOVE or edit_value is Ellipsis):
                     node.pop(edit_key)
                     edit_offset += 1
+                elif isinstance(node, list):
+                    node[edit_key] = edit_value
                 else:
-                    if isinstance(node, list):
-                        node[edit_key] = edit_value
-                    else:
-                        setattr(node, edit_key, edit_value)
+                    setattr(node, edit_key, edit_value)
 
             idx = stack.idx
             keys = stack.keys
--- a/src/graphql/subscription/map_async_iterator.py
+++ b/src/graphql/subscription/map_async_iterator.py
@@ -59,6 +59,7 @@ class MapAsyncIterator:
                 raise error
 
             value = anext.result()
+
         result = self.callback(value)
 
         return await result if isawaitable(result) else result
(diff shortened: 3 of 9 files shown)


Cito (commented):
Hi @yezz123, thanks for your contribution.

I'll go through your changes as time permits. One remark upfront: I try to keep the code very close to the original GraphQL.js code, so that the task of keeping the code bases in sync stays managable. That's why I often keep certain constructs or order of code blocks even if the code could be simplified or optimized.

src/graphql/language/block_string.py
```diff
@@ -38,7 +38,7 @@ def dedent_block_string_value(raw_string: str) -> str:
 
 def is_blank(s: str) -> bool:
     """Check whether string contains only space or tab characters."""
-    return all(c == " " or c == "\t" for c in s)
+    return all(c in [" ", "\t"] for c in s)
```
Cito: It's disputable what would be best for readability and performance here. Note that there are at least 5 alternatives that look good to me:
- `c in " \t"`
- `c == " " or c == "\t"`
- `c in [' ', '\t']`
- `c in (' ', '\t')`
- `c in {' ', '\t'}`

The first solution only works if you are sure that `c` is a single character. Note that creating a list, tuple or set has a certain cost.

Out of curiosity, I measured with `timeit` and found that in fact `set` is the fastest solution, even if you only have 2 options. I would have guessed that the effect would be only noticeable with more options. Note that you want to optimize both cases, that the condition is true and that it is false. With the set, you have always a constant time. The other solutions are slower when the condition is false, and become worse the more options you have.

src/graphql/language/visitor.py
```diff
@@ -183,16 +183,15 @@ def __init_subclass__(cls) -> None:
...
-                        raise TypeError(f"Invalid AST node kind: {kind}.")
+            if attr in ("enter", "leave") and kind:
+                name = snake_to_camel(kind) + "Node"
+                node_cls = getattr(ast, name, None)
+                if (
+                    not node_cls
+                    or not isinstance(node_cls, type)
+                    or not issubclass(node_cls, Node)
+                ):
+                    raise TypeError(f"Invalid AST node kind: {kind}.")
```
Cito, line 194: Yes, that's better.

yezz123:
Hi @Cito 👋 

I got your idea about keeping the same one related to GraphQL.js code Yes, my changes are not something that touches the core, just a pythonic one related to refactoring some functions that need to function well and in a fast way.
I run the Local test all of the functions work smoothly ✨

src/graphql/language/visitor.py
```diff
@@ -256,22 +255,18 @@ def visit(root: Node, visitor: Visitor) -> Any:
...
                     node.pop(edit_key)
                     edit_offset += 1
+                elif isinstance(node, list):
+                    node[edit_key] = edit_value
                 else:
-                    if isinstance(node, list):
-                        node[edit_key] = edit_value
-                    else:
-                        setattr(node, edit_key, edit_value)
+                    setattr(node, edit_key, edit_value)
```
Cito, line 269: These look good, too.

src/graphql/language/block_string.py
Cito: If we make that change, we should look if there are similar comparisons and change all of them.

src/graphql/subscription/map_async_iterator.py
```diff
@@ -61,7 +59,7 @@ async def __anext__(self) -> Any:
                 raise error
 
             value = anext.result()
-            result = self.callback(value)
+        result = self.callback(value)
```
Cito, line 63: Ok, maybe add a new line in front of this.

src/graphql/utilities/strip_ignored_characters.py
```diff
@@ -31,40 +31,7 @@ def strip_ignored_characters(source: Union[str, Source]) -> str:
...
-          Field description
-          """
-          bar: String
-        }
-
-    Becomes::
-
-        """Type description""" type Foo{"""Field description""" bar:String}
-    '''
+    """
```
Cito: Why did you remove the rest of the docstring?

src/graphql/utilities/strip_ignored_characters.py
```diff
@@ -79,9 +46,10 @@ def strip_ignored_characters(source: Union[str, Source]) -> str:
         # Also prevent case of non-punctuator token following by spread resulting
         # in invalid token (e.g.`1...` is invalid Float token).
         is_non_punctuator = not is_punctuator_token_kind(current_token.kind)
-        if was_last_added_token_non_punctuator:
-            if is_non_punctuator or current_token.kind == TokenKind.SPREAD:
-                stripped_body += " "
+        if was_last_added_token_non_punctuator and (
+            is_non_punctuator or current_token.kind == TokenKind.SPREAD
+        ):
+            stripped_body += " "
```
Cito, line 85: This was a literal translation from GraphQL.js, but it's ok we can change that.

src/graphql/validation/rules/lone_anonymous_operation.py
```diff
@@ -20,9 +20,8 @@ def __init__(self, context: ASTValidationContext):
 
     def enter_document(self, node: DocumentNode, *_args: Any) -> None:
         self.operation_count = sum(
-            1
+            isinstance(definition, OperationDefinitionNode)
```
Cito, line 23: This makes use of the fact that bool is an int subtype in Python and `1 == True`.

Yes, we can do that.

src/graphql/validation/rules/single_field_subscriptions.py
```diff
@@ -23,62 +23,63 @@ class SingleFieldSubscriptionsRule(ValidationRule):
...
-                    field_selection_lists = list(fields.values())
-                    extra_field_selection_lists = field_selection_lists[1:]
-                    extra_field_selection = [
-                        field
-                        for fields in extra_field_selection_lists
-                        for field in (
-                            fields
-                            if isinstance(fields, list)
-                            else [cast(FieldNode, fields)]
+        if node.operation != OperationType.SUBSCRIPTION:
```
Cito, line 26: You just want to avoid one level of indentation, right? Again, this was a literal translation from GraphQL.js, but ok, we can do that.

src/graphql/validation/validation_context.py
```diff
@@ -97,10 +97,12 @@ def report_error(self, error: GraphQLError) -> None:
     def get_fragment(self, name: str) -> Optional[FragmentDefinitionNode]:
         fragments = self._fragments
         if fragments is None:
-            fragments = {}
-            for statement in self.document.definitions:
-                if isinstance(statement, FragmentDefinitionNode):
-                    fragments[statement.name.value] = statement
+            fragments = {
```
Cito, line 100: Yes, a dict comprehension is definitely better here.

src/graphql/utilities/print_schema.py
```diff
@@ -108,10 +108,7 @@ def is_schema_of_common_names(schema: GraphQLSchema) -> bool:
         return False
 
     subscription_type = schema.subscription_type
-    if subscription_type and subscription_type.name != "Subscription":
-        return False
-
-    return True
+    return not subscription_type or subscription_type.name == "Subscription"
```
Cito, line 111: Again, literally from GraphQL.js. But yes, let's do it a bit more elegantly here.

Cito (changes requested):

yezz123:
Hey @Cito I just fix all that you recommend to add, I agree with you in all of the changes look good and logical, we don't need to break the logic of it also we need to keep it in a pythonic way 🚀✨.

You could run all the workflows before the merge

Cito (approved):
Thank you @yezz123

yezz123:
Happy to contribute to one of the things that i use daily 🚀✨

src/graphql/language/block_string.py
Cito: For the records, I found that using a regex was ~even faster~ slower than all other solutions.

PR #150, opened 2021-12-10
Cache node hashing

AlecRosenbaum:
This change adds basic caching for `Node.__hash__`. The primary goal of this change is to optimize validation. The benchmark I tested against is `tests/benchmarks/test_validation_gql.py`.

Locally, this benchmark goes from `7ms` to `4ms` with this change. I've also done some testing against an internal schema from our application, using a large query from our application instead of the introspection query. In that test, validation time dropped from `28ms` to `11ms` locally.

### DIFF
--- a/src/graphql/language/ast.py
+++ b/src/graphql/language/ast.py
@@ -259,6 +259,10 @@ class Node:
             self._hash = hash(tuple(getattr(self, key) for key in self.keys))
         return self._hash
 
+    def __setattr__(self, key, value):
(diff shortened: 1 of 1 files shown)


Cito:
Thanks, will check this out. One minor disadvantage ist that when you modify the AST, then you must also reset the cashed hash value. Or maybe we should do it automatically in the `__setattr__` method?

AlecRosenbaum:
Good idea. I get about the same results with an added `__setattr__` override that resets `_hash`.

PR #152, opened 2021-12-21
Fix evaluation for pseudo values

fugal-dy:
If the value is e. g. a lazy translation from Django's gettext_lazy,
the value should be stringified to be concatenated.


fugal-dy:
fixes #153

Cito (commented):
Thank you for sending this PR. This makes sense, as GraphQL-core already supports registering lazy string types with `register_description`.

Maybe we should convert `value` to `str` already at the beginning of the `print_block_string` function, like this `if not isinstance(value, str): value = str(value)`. Then in the following, we can be sure that the string operations like `in` and `startswith` work properly even if not implemented.

Cito:
Made the above minor changes and merged this to the main branch. Thank you @fugal-dy.

PR #159, opened 2022-01-27
Make errors and extensions optional

syrusakbary:
Just looking a bit on the codebase.
I think both errors and extensions should be optional.

### DIFF
--- a/src/graphql/execution/execute.py
+++ b/src/graphql/execution/execute.py
@@ -99,9 +99,9 @@ __all__ = [
 class FormattedExecutionResult(TypedDict, total=False):
     """Formatted execution result"""
 
-    errors: List[GraphQLFormattedError]
+    errors: Optional[List[GraphQLFormattedError]]
     data: Optional[Dict[str, Any]]
-    extensions: Dict[str, Any]
+    extensions: Optional[Dict[str, Any]]
 
 
 class ExecutionResult:
--- a/src/graphql/execution/execute.py
+++ b/src/graphql/execution/execute.py
@@ -99,8 +99,8 @@ __all__ = [
 class FormattedExecutionResult(TypedDict, total=False):
     """Formatted execution result"""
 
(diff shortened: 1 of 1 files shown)


src/graphql/execution/execute.py
```diff
@@ -99,9 +99,9 @@
 class FormattedExecutionResult(TypedDict, total=False):
     """Formatted execution result"""
 
-    errors: List[GraphQLFormattedError]
     data: Optional[Dict[str, Any]]
-    extensions: Dict[str, Any]
+    errors: Optional[List[GraphQLFormattedError]]
```
Cito, line 103: The formatted property now does not include the errors and extensions if they are None. This is exrpressed by the total=False parameter of the TypedDict. So they should never be None. They are also not null in the JS version.

Cito: Regarding the order, it corresponds to the JS version where errors come first in the formatted and also the unformatted result. In graphql-core we always used a named tuple with data as first item for the unformatted result, probably because it looked more reasonable. So we face a dilemma here. Either be inconsistent with the JS formatted result or with our own unformatted result. Not sure what's worse. What do you think?

Cito: By the way, if we change nullability, we should do the same with GraphQLFormattedError.

syrusakbary:
ping @Cito

Cito:
Hi @syrusakbary did you see my inline comments above?

syrusakbary:
Do you mean here in Github? I can't see any comments (perhaps you forgot to submit the review?)

syrusakbary:
(closed by mistake)

Cito:
@syrusakbary Sorry, yes, forgot to submit.

src/graphql/execution/execute.py
syrusakbary: Got it, closing!

PR #161, opened 2022-03-10
Mention poetry instead of pipenv

conao3:
Looking at the source tree, it appears that the `pipenv` system file does not exist and that it is assumed to use `poetry` instead.
Therefore, I change the description to use poetry.


Cito (approved):
Goot point, thank you. Sorry for the delay.

conao3:
Thanks!

PR #164, opened 2022-04-01
Represent serialized floats to python float precision

bennyweise:
Currently, the general string format used in the serialisation of a float value uses the default precision of 6, which leads to a loss in precision of large/long numbers. This PR changes the precision to use up to 16 places, which is approximately the level of precision of a float. This doesn't have any impact on the existing tests that don't have numbers with precision > 6, and the added assertion in the existing test fails without the change.

### DIFF
--- a/src/graphql/utilities/ast_from_value.py
+++ b/src/graphql/utilities/ast_from_value.py
@@ -117,7 +117,7 @@ def ast_from_value(value: Any, type_: GraphQLInputType) -> Optional[ValueNode]:
         if isinstance(serialized, int):
             return IntValueNode(value=f"{serialized:d}")
         if isinstance(serialized, float) and isfinite(serialized):
-            return FloatValueNode(value=f"{serialized:g}")
+            return FloatValueNode(value=f"{serialized:.16g}")
 
         if isinstance(serialized, str):
             # Enum types use Enum literals.
--- a/src/graphql/utilities/ast_from_value.py
+++ b/src/graphql/utilities/ast_from_value.py
@@ -117,7 +117,10 @@ def ast_from_value(value: Any, type_: GraphQLInputType) -> Optional[ValueNode]:
         if isinstance(serialized, int):
             return IntValueNode(value=f"{serialized:d}")
         if isinstance(serialized, float) and isfinite(serialized):
-            return FloatValueNode(value=f"{serialized:.16g}")
+            value = str(serialized)
+            if value.endswith('.0'):
+                value = value[:-2]
+            return FloatValueNode(value=value)
 
         if isinstance(serialized, str):
             # Enum types use Enum literals.
--- a/src/graphql/utilities/ast_from_value.py
+++ b/src/graphql/utilities/ast_from_value.py
@@ -115,7 +115,7 @@ def ast_from_value(value: Any, type_: GraphQLInputType) -> Optional[ValueNode]:
 
         # Python ints and floats correspond nicely to Int and Float values.
         if isinstance(serialized, int):
-            return IntValueNode(value=f"{serialized:d}")
+            return IntValueNode(value=str(serialized))
         if isinstance(serialized, float) and isfinite(serialized):
             value = str(serialized)
             if value.endswith('.0'):
--- a/tests/utilities/test_ast_from_value.py
+++ b/tests/utilities/test_ast_from_value.py
@@ -86,6 +86,10 @@ def describe_ast_from_value():
 
         assert ast_from_value(1e40, GraphQLFloat) == FloatValueNode(value="1e+40")
 
+        assert ast_from_value(123456789.0123456, GraphQLFloat) == FloatValueNode(
+            value="123456789.0123456"
+        )
+
     def converts_string_values_to_string_asts():
         assert ast_from_value("hello", GraphQLString) == StringValueNode(value="hello")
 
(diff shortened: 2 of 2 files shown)


src/graphql/utilities/ast_from_value.py
```diff
@@ -117,7 +117,7 @@ def ast_from_value(value: Any, type_: GraphQLInputType) -> Optional[ValueNode]:
         if isinstance(serialized, int):
             return IntValueNode(value=f"{serialized:d}")
         if isinstance(serialized, float) and isfinite(serialized):
-            return FloatValueNode(value=f"{serialized:g}")
+            return FloatValueNode(value=f"{serialized:.16g}")
```
Cito: Yes, this is better than the default 6 digits used by the `g` format type.

However, to have the maximum precision you would actually need to use `.17g`, which on the other hand can create ugly representations like for `1.1`. I also hate to have hardcoded precision, because there may be Python implementations with different precision of floats.

To solve this, we could make use of the float serialization algorithm built-in to `str()` since Python 3.1, that always uses the shortest representation that does not lose precision.

So we could simply set `value=str(serialized)` in both cases. I think the only reason why I did not do this was that `1.0` should be serialized as `1`, not  as `1.0`. To achieve this, maybe it suffices to do a `str(serialized).removesuffix('.0')`? Do I overlook anything? It seems the existing tests and your additional test would be happy with that change.

tests/utilities/test_ast_from_value.py
```diff
@@ -86,6 +86,10 @@ def converts_float_values_to_float_asts():
 
         assert ast_from_value(1e40, GraphQLFloat) == FloatValueNode(value="1e+40")
 
+        assert ast_from_value(123456789.0123456, GraphQLFloat) == FloatValueNode(
+            value="123456789.0123456"
+        )
+
```
Cito, line 92: Can you add another test for `1.1` so that we avoid the ugly representation that we would get with `.17f` format?

Cito (changes requested):
Thanks a lot for this PR! Please check my comments below.

src/graphql/utilities/ast_from_value.py
Cito: Actually we need to do it like this, because `removesuffix` exists only since Py 3.9, and we still support Py 3.6+.
```py
            value = str(serialized)
            if value.endswith('.0'):
                value = value[:-2]
            return FloatValueNode(value=value)
```

bennyweise: Thanks for the comments! Yes, I tried first with .17g but it broke existing tests, so I was just using .16 as a pragmatic way of passing existing tests. I think using the `str` function is a better approach. Is the serialization of `1.0` as `1` defined in the spec? Naively I would have assumed that `1.0` would still serialize as `1.0`

tests/utilities/test_ast_from_value.py
bennyweise: Will add that test, as well as a test that encodes the `1.0` -> `1` serialization expectation as well. With the change to `str` serialization, I cal also change this test to assert 17 digits of precision.

src/graphql/utilities/ast_from_value.py
Cito: `str` actually used the `.17g` format before that change in Py 3.1 (which has been backported to Py 2.7 as well).

The serialization as `1.0` would be probably not against the spec, but I want to make it as compatible to the reference implementation in GraphQL.js as possible. There, `1.0` is serialized as `1`, probably because in JavaScript, the same type is used for floats and ints. There is already a test that checks `123.0` is serialized as `123` which has been directly taken over from the GraphQL.js test suite.

bennyweise: Ahh, I overlooked that in the test. I'll remove the superfluous additional test then.

tests/utilities/test_ast_from_value.py
```diff
@@ -86,6 +86,14 @@ def converts_float_values_to_float_asts():
 
         assert ast_from_value(1e40, GraphQLFloat) == FloatValueNode(value="1e+40")
 
+        assert ast_from_value(123456789.01234567, GraphQLFloat) == FloatValueNode(
+            value="123456789.01234567"
+        )
+
+        assert ast_from_value(1.0, GraphQLFloat) == FloatValueNode(value="1")
```
Cito: We already have a similar test above, but it will not harm adding this one as well.

src/graphql/utilities/ast_from_value.py
```diff
@@ -117,7 +117,10 @@ def ast_from_value(value: Any, type_: GraphQLInputType) -> Optional[ValueNode]:
         if isinstance(serialized, int):
             return IntValueNode(value=f"{serialized:d}")
```
Cito: I think we should then set `value=str(serialized)` here as well, just for symmetry with the case below.

Cito (changes requested):
Looks good. Can you change the IntValueNode case to use `str` as well? Should make no difference, but the code will look more consistent then.

src/graphql/utilities/ast_from_value.py
bennyweise: Makes sense, I've now made that change.

Cito (approved):

PR #168, opened 2022-04-25
Exclude unset fields

dima-dmytruk23:
Original discussion: https://github.com/graphql-python/graphene-pydantic/pull/75

I use `graphene` and `graphene-pydantic` libraries.
Code example:
https://gist.github.com/dima-dmytruk23/aaeba0fbc7a539c1f8bf3d0914fce580

> The client does not pass the name field, but it is still present in the `mutation` as `None`. Input query turns into a `UserUpdateInput`, which is when the default values are filled in to the dictionary. So then when your code passes the dictionary in to build the `UserUpdate`, it sets all the fields -- so `exclude_unset` doesn't exclude anything, since all the fields were in fact set.
> 
> I am fairly sure it's not in `graphene-pydantic`, though, since that is only responsible for converting to the `GrapheneInputObjectType`.

I propose to resolve this issue by adding the `exclude unset` flag to the `GraphQLSchema` class and use it in the `coerce_input_value` function.

### DIFF
--- a/src/graphql/execution/values.py
+++ b/src/graphql/execution/values.py
@@ -136,7 +136,7 @@ def coerce_variable_values(
             )
 
         coerced_values[var_name] = coerce_input_value(
-            value, var_type, on_input_value_error
+            value, var_type, on_input_value_error, exclude_unset=schema.exclude_unset
         )
 
     return coerced_values
--- a/src/graphql/type/schema.py
+++ b/src/graphql/type/schema.py
@@ -342,6 +342,7 @@ class GraphQLSchema:
             ast_node=deepcopy(self.ast_node),
             extension_ast_nodes=deepcopy(self.extension_ast_nodes),
             assume_valid=True,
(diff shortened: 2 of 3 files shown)


Cito:
Hi @dima-dmytruk23, and thanks for contributing.

Please understand that the scope of GraphQL-core is to be a straight-forward port of GraphQL.js, so I am very reluctant to add features that are not part of GraphQL.js or change the API. I'm not saying that we can never do this, and we in fact have serveral such extensions, but you really need to have convincing reasons why this is necessary. Any such extension makes maintaining GraphQL-core and keeping it in sync with GraphQL.js more difficult as they sum up.

Therefore, could you please first open an issue and explain what the feature tries to achieve, and how it could be useful and why it cannot be achieved in other ways, with example code, independent of Graphene or Graphene-Pydantic? We also require unit tests, and this will help to create them.

Cito:
See issue #170.

Issue #153, opened 2021-12-21
Printing block string fails for lazy evaluated values


fugal-dy:
using graphql-core in the context of Django projects `print_block_string` could receive a proxy object from a lazy translation.

In this case it fails to concatenate the value.

Expected: evaluate the proxy object in place as a string

Cito:
This is now in the main branch and will be available soon in v3.2.

PR #175, opened 2022-06-08
Bugfix: Cast error message to string to handle proxy objects

jkimbo:
Some frameworks (Django as an example) use a lazy evaluated object to represent error messages (https://github.com/django/django/blob/49b470b9187b6be60e573fed08c8f4a87f133750/django/utils/functional.py#L87-L96). This breaks graphql-core when it tries to format the error as a string. 

This PR fixes this bug by forcing the error message to a string when creating a GraphQLError object.

### DIFF
--- a/src/graphql/error/located_error.py
+++ b/src/graphql/error/located_error.py
@@ -29,7 +29,7 @@ def located_error(
         return original_error
     try:
         # noinspection PyUnresolvedReferences
-        message = original_error.message  # type: ignore
+        message = str(original_error.message)  # type: ignore
     except AttributeError:
         message = str(original_error)
     try:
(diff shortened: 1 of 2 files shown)


jkimbo:
I changed my approach to this fix by forcing the error message to a string in the `located_error` function instead of the `GraphQLError.__str__` method.

Cito:
Thank you @jkimbo, will look into this later. I think #153 was something similar.

Cito (approved):
Looks good. In addabc096e5ce7f1f787d58c03788114a9eda2cf I have simplified the test a bit and made it more similar to the other one in `test_block_string`.

jkimbo:
Thanks @Cito

PR #179, opened 2022-09-05
pyproject.toml: description must be one line

singingwolfboy:
See https://github.com/python-poetry/poetry/issues/3610 and https://packaging.python.org/en/latest/specifications/core-metadata/#summary


Cito:
Thank you for the contribution @singingwolfboy and sorry for the delay. This is actually the same as #180 which was already merged, so closing this.

PR #181, opened 2022-09-27
Support custom execution contexts in subscriptions

bkad:
This adds support for the `execution_context_class` argument to `subscribe` which already exists for `execute`. My usecase had an existing need for a custom ExecutionContext for queries, so I was surprised when trying to add subscriptions that it wasn't already supported.


Cito (approved):
Nice.

Cito:
Thanks a lot for your contribution, @bkad.

PR #182, opened 2022-10-20
sort enums by name when printing schema

dctalbot:
We run the following to generate a schema file, but the enums are not printed in a reliable order. This creates unnecessary code diffs.

```py
from graphql.utils import schema_printer
from .gql import schema

schema_str = schema_printer.print_schema(schema)
fp = open("gql/schema.graphql", "w")
fp.write(schema_str)
fp.close()
```

This PR attempts to solve this by sorting the enum types by name before printing.

### DIFF
--- a/src/graphql/utilities/print_schema.py
+++ b/src/graphql/utilities/print_schema.py
@@ -180,7 +180,7 @@ def print_enum(type_: GraphQLEnumType) -> str:
         print_description(value, "  ", not i)
         + f"  {name}"
         + print_deprecated(value.deprecation_reason)
-        for i, (name, value) in enumerate(type_.values.items())
+        for i, (name, value) in enumerate(sorted(type_.values.items(), key=lambda x: x[0]))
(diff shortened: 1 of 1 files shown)


Cito:
You can already use `lexicographic_sort_schema` to get a sorted version of the schema, and print that, if you need a reliable order. This would also solve the same problem with other structures.

If you want to argue that the enum values should also be sorted in the standard print function, then I prefer that you suggest this in a [GraphQL.js](https://github.com/graphql/graphql-js) issue. If it's a good idea, it will be implemented there, and then ported here. That's the way how new features or changed behavior is implemented in GraphQL-core. Currently the values are not sorted in GraphQL.js - see [here](https://github.com/graphql/graphql-js/blob/f851eba93167b04d6be1373ff27927b16352e202/src/utilities/printSchema.ts#L194) and the goal of GraphQL-core is simply to copy the behavior of GraphQL.js.

PR #194, opened 2023-03-02
BugFix: Nested List Input Coercion not matching GraphQL spec

abhinand-c:
This PR fixes the issue of Nested List Input with multiple values not being properly validated as per the schema.

For a nested List ` [[Int!]] `, providing an input value of `[1, 2, 3] `
**[Expected Behaviour](https://spec.graphql.org/October2021/#sel-HAHjBJLDLEBAAADFAb0le):** should raise validation error
**Current Behaviour:** coerce value to `[ [1], [2], [3] ]`

### Changes
- Adds check to validate and raises error for the condition
- Updated tests to reflect the new logic, and validate error

### Refer: 
- https://spec.graphql.org/October2021/#sec-List.Input-Coercion

image

### DIFF
--- a/src/graphql/utilities/coerce_input_value.py
+++ b/src/graphql/utilities/coerce_input_value.py
@@ -70,7 +70,15 @@ def coerce_input_value(
         if is_iterable(input_value):
             coerced_list: List[Any] = []
             append_item = coerced_list.append
+            is_nested_list = bool(is_list_type(item_type) and len(input_value) > 1)
             for index, item_value in enumerate(input_value):
+                if is_nested_list and not is_iterable(item_value):
+                    # All input values should be iterable for multivalued nested list types
+                    on_error(
+                        path.as_list() if path else Path(path, index, None).as_list(),
+                        item_value,
+                        GraphQLError(f"Expected type '{inspect(item_type)}' to be a list."),
+                    )
                 append_item(
                     coerce_input_value(
                         item_value, item_type, on_error, Path(path, index, None)
--- a/src/graphql/utilities/coerce_input_value.py
+++ b/src/graphql/utilities/coerce_input_value.py
@@ -70,14 +70,18 @@ def coerce_input_value(
         if is_iterable(input_value):
             coerced_list: List[Any] = []
             append_item = coerced_list.append
-            is_nested_list = bool(is_list_type(item_type) and len(input_value) > 1)
+            is_nested_list = bool(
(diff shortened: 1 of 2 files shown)


Cito:
Hi @abhinand-c , and thank you for reporting and providing a fix. May I suggest that next time you create an issue before posting a PR ? We can then discuss what's the best way to solve the problem before you invest too much time.

I believe you're right in that the coercion rule (coercion needs to be distinguished from validation) is not properly implemented according to the latest spec. The implementation was correct regarding the older spec of 2016, but it looks like in 2018 the spec was [changed](https://github.com/graphql/graphql-spec/pull/436/files), and this change was not incorporated into GraphQL.js of which GraphQL-core is a port.

The proper procedure to fix this is to open an issue for GraphQL.js, maybe followed by a PR with a fix for TypeScript, and when this is merged into GraphQL.js, the same fix will be ported to Python. By following this procedure we make sure the code base does not deviate, the JavaScript users benefit as well, and we can have a discussion with the larger JavaScript community and GraphQL experts on what's the best way to fix it and review the PR.

Let me know if you want to take care of that, or if you want me to do it and if I can take your code as a base for a possible PR in GraphQL.js.

abhinand-c:
Hi @Cito , thanks for the guiding me in the right way, I'll definitely follow the proper procedure next time. I have raised the issue in GraphQL.js (https://github.com/graphql/graphql-js/issues/3858). 

Also, currently in this PR my commits were not signed, shall I proceed to update my commits and sign them? 
Is there anything else that I should have taken care of as a first time contributor here?

Cito:
No worries @abhinand-c it's all well. GraphQL.js is our upstream project. The guidelines for contributing to GraphQL.js are [these](https://github.com/graphql/graphql-js/blob/main/.github/CONTRIBUTING.md) if you want to create a PR there, but here we are a bit more relaxed.

abhinand-c:
Hi, I have updated old (unsigned) commits and replaced them with signed commits by force pushing.

erikwrede:
Upstream update: https://github.com/graphql/graphql-spec/pull/1057#issuecomment-2392084498

Once the spec PR is merged, we can close this PR, as the table this relied on was adjusted to allow our explicit case.

abhinand-c:
Closing this PR as spec was updated to clarify the isssue, and no change in implementation is required.  

**Ref:** 
- https://github.com/graphql/graphql-js/issues/3858#issuecomment-2495638447
- https://github.com/graphql/graphql-spec/pull/1057

PR #197, opened 2023-04-21
Simplify MapAsyncIterable using async generator semantics

kristjanvalur:
This PR rewrites the `MapAsyncIteable` using simple native async primitives

## Problem

The existing implementation of `MapAsyncIterable` has several problems, mostly to do with `aclose()` semantics.
This has become evident in projects such as [Strawberry-graphql](https://github.com/strawberry-graphql/strawberry)
where calling `aclose()` on a subscription sometimes behaves in unexpected manner.  Fundamentally this is caused by the unnecessarily complex implementation, containing inner `Task` objects (created by `ensure_future()`) created for each iterated value.

A concrete problem which sometimes manifests itself is that calling `aclose()` on this iterator results in a RuntimeError.  The reason is that the actual `__anext()__` call is currently performed in a separate `Task`.  Even when that task is cancelled it may still be pending

The iterable _tries_ to look like an `AsyncGenerator` but fails to correctly implement the changes of `aclose()` which
added in Python 3.8, where such an iterator is not allowed to be re-entrant.

Ultimately, the class-based implementation is unnecessarily complex for something as fundamentally simple as iterating over, and mapping, another iterable.

## Changes

This PR rewrites this iterator using an AsyncGenerator, thus ensuring that simple and understandable internal mechanisms
are used.  It avoids any use of synchronization primitives and Task objects and simply iterates over the results.  It provides the extended `aclose()` optional semantics for the inner iterable as expected.

We remove unnecessary unit tests for this class and fix the ones making strange assumptions, such as ones making it allowable to ignore GeneratorExits and _yield_ data.  Also a test for un-closing an iterator is removed, since that is a completely un-pythonic idea.

## Further improvements

Further enhancements might be:

- remove the `athrow()` and `is_closed` members which serve no purpose
- implement this method entirely as an Async Generator, and drop the special type, which only complicates code.  The relevant checks in graphql-core could then use the `inspect.isasyncgen()` method instead.

### DIFF
--- a/src/graphql/execution/map_async_iterable.py
+++ b/src/graphql/execution/map_async_iterable.py
@@ -1,8 +1,7 @@
 from __future__ import annotations  # Python < 3.10
 
-from inspect import isawaitable
 from types import TracebackType
-from typing import Any, AsyncIterable, Callable, Optional, Type, Union
+from typing import Any, AsyncIterable, Awaitable, Callable, Optional, Type, Union
 
 
 __all__ = ["MapAsyncIterable"]
@@ -11,18 +10,21 @@ __all__ = ["MapAsyncIterable"]
 # The following is a class because its type is checked in the code.
 # otherwise, it could be implemented as a simple async generator function
 
+
 # noinspection PyAttributeOutsideInit
 class MapAsyncIterable:
     """Map an AsyncIterable over a callback function.
 
     Given an AsyncIterable and a callback function, return an AsyncIterator which
-    produces values mapped via calling the callback function.
+    produces values mapped via calling the callback async function.
 
-    When the resulting AsyncIterator is closed, the underlying AsyncIterable will also
-    be closed.
+    Similar to an AsyncGenerator, an `aclose()` method is provivde which
+    will close the underlying AsyncIterable be if it has an `aclose()` method.
     """
 
-    def __init__(self, iterable: AsyncIterable, callback: Callable) -> None:
+    def __init__(
+        self, iterable: AsyncIterable[Any], callback: Callable[[Any], Awaitable[Any]]
+    ) -> None:
         self.iterator = iterable.__aiter__()
         self.callback = callback
         self._ageniter = self._agen()
@@ -39,8 +41,7 @@ class MapAsyncIterable:
     async def _agen(self) -> Any:
         try:
             async for v in self.iterator:
-                result = self.callback(v)
-                yield (await result) if isawaitable(result) else result
+                yield await self.callback(v)
         finally:
             self.is_closed = True
             if hasattr(self.iterator, "aclose"):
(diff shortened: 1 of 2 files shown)


Cito:
Thanks a lot @kristjanvalur. I always wanted to revisit this part of the code anyway. Will definitely look into this, but please have some patience since I'm currently busy with other things.

kristjanvalur:
Also,  I think it is prudent for you to review the semantics here.  I see that this has recently been renamed to "Iterable" rather than "Iterator".    I am not sure that this is wise.  A list is an "iterable", but a list iterator is, while technically also an iterable, is an iterator.   The concept of an iterable is something which can be iterated over multiple times.
In the sense of subscriptions, this makes no sense.   (You don't want to execute the same subscription multiple times)
If something has a `__next__()` method, it is primarily an iterator. (same goes for async)

kristjanvalur:
By the way, one of the bad side effects of the previous implementation is that it makes debugging hard.  The `__anext__()` method is executed in its own `Task` which means that the call chain is broken.

Cito:
> Also, I think it is prudent for you to review the semantics here. I see that this has recently been renamed to "Iterable" rather than "Iterator". I am not sure that this is wise.

I was mirroring [that change in Change.js](https://github.com/graphql/graphql-js/commit/4822cb4bedc7bc45d6c4be136e9180fb0312a9bb) here. I think JavaScript was inspired by Python regarding iterables, so the semantics is pretty similar. The reason for the renaming was that the class has an `__aiter__` method. So it is not *only* an Iterator, but also an Iterable (similar for JavaScript).

Remember that the goal of GraphQL.js is to be a pretty much 1:1 translation of GraphQL.js to Python. I try to avoid doing or naming things differently, because only by staying close to the original it is sustainable to maintain the library and keep it up to date with the current developments.

For instance, I'm currently working on getting the `@defer/@stream` directives from GraphQL.js into GraphQL.core which is quite a big feature which requires many changes in the execution module. I will look into your PR once I'm done with that big change, and make sure it's compatible.

> By the way, one of the bad side effects of the previous implementation is that it makes debugging hard. The `__anext__()` method is executed in its own Task which means that the call chain is broken.

Agreed, that's another good reason to change it.

kristjanvalur:
Yes, no rush at all.  I'll probably implement workarounds in strawberry anyway.  

I also had time to think it over and I don't think my comments on iterables/iterators are necessarily correct, I guess I got caught up in the moment.  Cheers!

kristjanvalur:
Closing this in favor of #199 , now that https://github.com/graphql-python/graphql-core/commit/62749e5a397946ef6e8186e759360aad3cb7715b has been committed.  There is no need to keep the MapAsyncIterable _class_ around.

Cito:
Thanks. Will try to get #199 into the next pre-release (very soon).

PR #203, opened 2023-06-09
bugfix Update definition.py

evanmays:
Looks like small typo


Cito (approved):
Thank you, well spotted.

Issue #213, opened 2024-02-20
Add the ability to specify `GraphQLResolveInfo.context` type


fedirz:
Currently, `GraphQLResolveInfo.context` is being typed as `Any`, making `GraphQLResolveInfo` a generic would give users the ability to provide their type. This is already being done in `GraphQL.js`
```
export type GraphQLFieldResolver<
  TSource,
  TContext,
  TArgs = any,
  TResult = unknown,
> = (
  source: TSource,
  args: TArgs,
  context: TContext,
  info: GraphQLResolveInfo,
) => TResult;
```

If this change will be welcome, I'd like to make a pull-request implementing it.

Cito:
Sure, PRs of that kind are always welcome.

PR #214, opened 2024-02-20
Allow user to pass in a custom resolve info context type (#213)

fedirz:
Usage example:
```
from typing import Generic, NamedTuple, TypeVar

TContext = TypeVar("TContext")

class GraphQLResolveInfo(Generic[TContext], NamedTuple):
    ...
    context: TContext
    ...

GraphQLResolveInfo(None) # Ok
GraphQLResolveInfo({"asdf": 1234}) # Ok

type Context = dict[str, int]

UserGraphQLResolveInfo = GraphQLResolveInfo[Context]

UserGraphQLResolveInfo(None)
 # │     Argument of type "None" cannot be assigned to parameter "context" of type "Context" in function "__init__" Pyright (reportArgumentType) [17, 24]
 # │        "None" is incompatible with "Context"

UserGraphQLResolveInfo({"asdf": 1234}) # Ok
```

### DIFF
--- a/src/graphql/type/definition.py
+++ b/src/graphql/type/definition.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations  # Python < 3.10
 
+import sys
 from enum import Enum
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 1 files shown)


src/graphql/type/definition.py
Cito, line 1: Shouldn't the context attribute now get the type TContext instead of Any?

src/graphql/type/definition.py
```diff
@@ -554,8 +554,9 @@ def to_kwargs(self) -> GraphQLFieldKwargs:
     def __copy__(self) -> GraphQLField:  # pragma: no cover
         return self.__class__(**self.to_kwargs())
 
+TContext = TypeVar("TContext")
 
-class GraphQLResolveInfo(NamedTuple):
+class GraphQLResolveInfo(Generic[TContext], NamedTuple):
```
Cito:
```suggestion
class GraphQLResolveInfo(NamedTuple, Generic[TContext]):
```

I think it's nicer to always put the Generics at the end.

Cito (changes requested):
Can you also add a test that serves as a usage example, like above in the description.

You can simulate a type error with a "type: ignore" comment. It should say "unnecessary type ignore" if there is no type error.

src/graphql/type/definition.py
Cito: I also noticed that multiple inheritance with NamedTuple is not supported with Python < 3.11 (see [here](https://github.com/python/cpython/issues/88089)). Since GraphQL-core supports Python > 3.6, we need to make a case distinction or catch the type error for older Python versions, and use a normal NamedTuple in that case.

Cito:
How did you actually produce that type error? I tried with mypy 1.8, but couldn't reproduce it.

Update: I can reproduce it now, but not if I use a type alias. I need to specify the type directly. Seems you used Pylance, right?

fedirz:
> How did you actually produce that type error? I tried with mypy 1.8, but couldn't reproduce it.
> 
> Update: I can reproduce it now, but not if I use a type alias. I need to specify the type directly. Seems you used Pylance, right?

No, the error came from `pyright`
```
graphql-core on  generic-resolve-info-context [!?] is 📦 v3.3.0a4
❄ ❯ python --version
Python 3.12.2

graphql-core on  generic-resolve-info-context [!?] is 📦 v3.3.0a4
❄ ❯ pyright --version
pyright 1.1.349

graphql-core on  generic-resolve-info-context [!?] is 📦 v3.3.0a4
❄ ❯ pyright src/graphql/test.py
/home/ted/code/graphql-core-dev/graphql-core/src/graphql/test.py
  /home/ted/code/graphql-core-dev/graphql-core/src/graphql/test.py:17:24 - error: Argument of type "None" cannot be assigned to parameter "context" of type "Context" in function "__init__"
    "None" is incompatible with "Context" (reportArgumentType)
1 error, 0 warnings, 0 informations
```

fedirz:
> Can you also add a test that serves as a usage example, like above in the description.
> 
> You can simulate a type error with a "type: ignore" comment. It should say "unnecessary type ignore" if there is no type error.

Is there already a test in the codebase that does something similar, which I can use as an example?

Cito:
> > How did you actually produce that type error? I tried with mypy 1.8, but couldn't reproduce it.
> > Update: I can reproduce it now, but not if I use a type alias. I need to specify the type directly. Seems you used Pylance, right?
> 
> No, the error came from `pyright`

Makes sense since Pylance is based on Pyright.

Cito:
The idea is to add a test that serves as a usage example and where running the test or type checking of the test code would fail if you revert the change in this PR.

Instead of a comment like "# this gives an error" you would simply write "# type: ignore". This comment tells type checkers like pyright or mypy to ignore the typing error, so that the type checking that is also run on the test code would not fail. On the other hand, if there is *not* the expected type error, mypy is configured to report a superfluous "type: ignore" comment. You can test everything with tox and the included tox.ini file.

But you can also leave the tests to me, no problem.

fedirz:
> The idea is to add a test that serves as a usage example and where running the test or type checking of the test code would fail if you revert the change in this PR.
> 
> Instead of a comment like "# this gives an error" you would simply write "# type: ignore". This comment tells type checkers like pyright or mypy to ignore the typing error, so that the type checking that is also run on the test code would not fail. On the other hand, if there is _not_ the expected type error, mypy is configured to report a superfluous "type: ignore" comment. You can test everything with tox and the included tox.ini file.
> 
> But you can also leave the tests to me, no problem.

Yeah, if you could add the test, that'd be great. Thanks for the help on this!

Cito (approved):

PR #218, opened 2024-04-09
Enable recursive type definitions

yilinjuang:
Mypy supports recursive types now.

https://github.com/python/mypy/issues/731
https://github.com/python/mypy/issues/14219

### DIFF
--- a/src/graphql/pyutils/path.py
+++ b/src/graphql/pyutils/path.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, NamedTuple, Optional
+from typing import NamedTuple
 
 __all__ = ["Path"]
 
@@ -10,7 +10,7 @@ __all__ = ["Path"]
 class Path(NamedTuple):
     """A generic path of string or integer indices"""
 
-    prev: Optional[Path]
+    prev: Path | None
     """path with the previous indices"""
     key: str | int
     """current index in the path (string or integer)"""
@@ -25,7 +25,7 @@ class Path(NamedTuple):
         """Return a list of the path keys."""
         flattened: list[str | int] = []
         append = flattened.append
-        curr: Path = self
+        curr: Path | None = self
         while curr:
             append(curr.key)
             curr = curr.prev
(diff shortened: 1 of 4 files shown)


src/graphql/pyutils/path.py
```diff
@@ -2,15 +2,15 @@
...
+from typing import Any, NamedTuple, Optional
 
 __all__ = ["Path"]
 
 
 class Path(NamedTuple):
     """A generic path of string or integer indices"""
 
-    prev: Any  # Optional['Path'] (python/mypy/issues/731)
+    prev: Optional[Path]
```
Cito:
```suggestion
    prev: Path |  None
```

recently switched to this newer style

src/graphql/validation/rules/overlapping_fields_can_be_merged.py
```diff
@@ -98,10 +98,7 @@ def enter_selection_set(self, selection_set: SelectionSetNode, *_args: Any) -> N
 # Field name and reason.
 ConflictReason: TypeAlias = Tuple[str, "ConflictReasonMessage"]
 # Reason is a string, or a nested list of conflicts.
-if MYPY:  # recursive types not fully supported yet (/python/mypy/issues/731)
```
Cito, line 101: You can now remove the MYPY constant completely.

src/graphql/pyutils/path.py
Cito: Line 28 must then also be changed to `curr: Path | None = self`.

Also, in `test_executor` we must now add the line `assert prev is not None` two times (as lines 311, 315).

Cito (approved):
Thanks you for contributing.

There are still some minor issues, see failing QA test and comments.

src/graphql/validation/rules/overlapping_fields_can_be_merged.py
yilinjuang: Removed

src/graphql/pyutils/path.py
yilinjuang: Fixed

PR #227, opened 2024-09-22
add `ValidationAbortedError` to `__all__`

trim21:


Cito (approved):
GraphQL.js doesn't export it either, but I think you're right, it cannot harm to export it.

Cito:
It will not be part of the official, documented API though, and not available directly from the graphql package to end users. If you want that, please raise it as an issue in GraphQL.js so that we can stay compatible.

PR #230, opened 2024-10-19
feat: add codspeed for continuous benchmarking

erikwrede:
This PR adds CodSpeed for benchmarking


tests/benchmarks/test_visit.py
```diff
@@ -23,5 +23,5 @@ def test_visit_all_ast_nodes(benchmark, big_schema_sdl):  # noqa: F811
 def test_visit_all_ast_nodes_in_parallel(benchmark, big_schema_sdl):  # noqa: F811
     document_ast = parse(big_schema_sdl)
     visitor = DummyVisitor()
-    parallel_visitor = ParallelVisitor([visitor] * 50)
+    parallel_visitor = ParallelVisitor([visitor] * 20)
```
Cito, line 26: Is there a reason why you reduced the number? The original benchmark had 50.

.github/workflows/test.yml
```diff
@@ -3,6 +3,29 @@ name: Tests
...
 
 jobs:
+
+  benchmarks:
+    name: 📈 Benchmarks
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - run: pipx install poetry
```
Cito: Shouldn't this come after the setup-python?

Cito (approved):
Thanks, looks good. I have made two comments. I'd also move it into a separate yml file.

You can ignore the mypy issues since they are not related to the PR, I fill fix them in main.

tests/benchmarks/test_visit.py
erikwrede: Benchmark timed out on CI with 50 visitors. Couldn't reproduce, so I reduced the number and now it's working perfectly fine.

.github/workflows/test.yml
erikwrede: Good point, I honestly copied this job over from strawberry, where' it's been running smoothly for a while. I can change the order if you prefer.

tests/benchmarks/test_visit.py
Cito: Ok, just wanted to know. I'll add a comment.

.github/workflows/test.yml
Cito: Saw you did that already, thanks.

PR #231, opened 2024-10-21
fix: recreate poetry lock

erikwrede:
Recreating the poetry lock, because poetry install failed on master


erikwrede:
The recent master merge seems to have broken it ^

poetry.lock
```diff
@@ -1,10 +1,9 @@
-# This file is automatically @generated by Poetry 1.4.2 and should not be changed by hand.
```
erikwrede, line 1: Seems like our poetry versions differ, most of the changes here should be caused by that. It was a simple `poetry lock --no-update`

Cito (approved):

PR #234, opened 2025-02-16
Fix `IntrospectionQuery` type definition

sobolevn:
`TypedDict` mangles private names as regular Python classes do. So, `__schema` name will become `_IntrospectionQuery__schema` instead. But, inline definition will preserve a correct name for the field. See https://github.com/python/cpython/issues/129567

Found during https://github.com/python/mypy/pull/16715

### DIFF
--- a/src/graphql/utilities/get_introspection_query.py
+++ b/src/graphql/utilities/get_introspection_query.py
@@ -302,7 +302,5 @@ class IntrospectionSchema(MaybeWithDescription):
     directives: list[IntrospectionDirective]
 
 
-class IntrospectionQuery(TypedDict):
-    """The root typed dictionary for schema introspections."""
-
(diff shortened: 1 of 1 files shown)


Cito:
Thanks for reporting and the PR. Is this also a problem with current mypy versions? If yes, how can I provoke a typing error there?

sobolevn:
Not until https://github.com/python/mypy/pull/16715 is merged

Cito:
Looks good, but you need to silence the linter with ` # noqa: UP013` because it prefers class syntax, and add a small comment as explanation why we deviate from the class syntax here.

Cito:
Now there are only mypy issues, probably because mypy has been updated and not related to this PR. Will fix them separately.

PR #237, opened 2025-05-03
Update README.md to include Typed GraphQL

willemt:
Hi there, I maintain a small library that uses GraphQL-Core so was wondering if it was OK to include in the "Integrations" section? Thanks for GraphQL-Core and your time


Cito (approved):

Cito:
Thank you, looks interesting.

PR #238, opened 2025-05-13
Cancel remaining fields on exceptions

### LINKED ISSUE #236: `RuntimeWarning: coroutine ... never awaited` when a sibling resolver raises an error
**Repro**

```python
import asyncio
from graphql import (
    graphql,
    GraphQLSchema,
    GraphQLObjectType,
    GraphQLField,
    GraphQLNonNull,
    GraphQLString,
)


async def resolve_foo(obj, info):
    return "foo!"


def resolve_bar(obj, info):
    raise Exception('yikes')


schema = GraphQLSchema(
    query=GraphQLObjectType(
        name="RootQueryType",
        fields={
            "foo": GraphQLField(GraphQLNonNull(GraphQLString), resolve=resolve_foo),
            "bar": GraphQLField(GraphQLNonNull(GraphQLString), resolve=resolve_bar),
        },
    )
)

result = asyncio.run(graphql(schema, "{ foo bar }"))
print(result)
```

Prints

```
$ python3 test2.py
ExecutionResult(data=None, errors=[GraphQLError('yikes', locations=[SourceLocation(line=1, column=7)], path=['bar'])])
sys:1: RuntimeWarning: coroutine 'ExecutionContext.execute_field.<locals>.await_result' was never awaited
sys:1: RuntimeWarning: coroutine 'resolve_foo' was never awaited
```

**Version info**

```
$ pip freeze | grep graphql-core
graphql-core==3.2.6
```

I'm guessing what could be happening here is that:

- non-nullable field error (`bar`) bubbles up and cancels execution for the rest of the type's fields
- ...but we already kicked off the `foo` async resolver function
- ...but by aborting, we don't ever unwrap the future / await the result for, hence the warning.

Which makes total sense -- and indeed if we do `schema.execute("{ bar foo }")` we don't see the warning printed -- because the async function never even gets kicked off.

I think maybe we need to be calling `task.cancel()` or `.close()` somewhere?

- https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel
- https://docs.python.org/3/reference/datamodel.html#coroutine.close

(originally misatttributed to strawberry https://github.com/strawberry-graphql/strawberry/issues/3820) 

Thanks!

mgorven:
gather() returns when the first exception is raised, but does not cancel any remaining tasks. These continue to run which is inefficient, and can also cause problems if they access shared resources like database connections.

Fixes: #236

### DIFF
--- a/src/graphql/execution/execute.py
+++ b/src/graphql/execution/execute.py
@@ -2,7 +2,14 @@
 
 from __future__ import annotations
 
-from asyncio import ensure_future, gather, shield, wait_for
+from asyncio import (
+    CancelledError,
+    create_task,
+    ensure_future,
+    gather,
+    shield,
+    wait_for,
+)
 from contextlib import suppress
 from copy import copy
 from typing import (
@@ -459,12 +466,23 @@ class ExecutionContext:
                 field = awaitable_fields[0]
                 results[field] = await results[field]
             else:
-                results.update(
-                    zip(
-                        awaitable_fields,
-                        await gather(*(results[field] for field in awaitable_fields)),
-                    )
-                )
+                tasks = {
+                    create_task(results[field]): field  # type: ignore[arg-type]
+                    for field in awaitable_fields
+                }
+
+                try:
+                    awaited_results = await gather(*tasks)
+                except Exception:
+                    # Cancel unfinished tasks before raising the exception
+                    for task in tasks:
+                        if not task.done():
+                            task.cancel()
+                    await gather(*tasks, return_exceptions=True)
+                    raise
+
+                results.update(zip(awaitable_fields, awaited_results))
+
             return results
 
         return get_results()
@@ -538,6 +556,10 @@ class ExecutionContext:
                     try:
                         return await completed
                     except Exception as raw_error:
+                        # Before Python 3.8 CancelledError inherits Exception and
+                        # so gets caught here.
+                        if isinstance(raw_error, CancelledError):
+                            raise  # pragma: no cover
                         self.handle_field_error(
                             raw_error,
                             return_type,
@@ -745,6 +767,10 @@ class ExecutionContext:
             if self.is_awaitable(completed):
                 completed = await completed
         except Exception as raw_error:
+            # Before Python 3.8 CancelledError inherits Exception and
+            # so gets caught here.
+            if isinstance(raw_error, CancelledError):
+                raise  # pragma: no cover
             self.handle_field_error(
                 raw_error, return_type, field_group, path, incremental_data_record
             )
--- a/tests/execution/test_parallel.py
+++ b/tests/execution/test_parallel.py
@@ -11,6 +11,7 @@ from graphql.type import (
     GraphQLInt,
     GraphQLInterfaceType,
     GraphQLList,
+    GraphQLNonNull,
     GraphQLObjectType,
     GraphQLSchema,
     GraphQLString,
@@ -193,3 +194,42 @@ def describe_parallel_execution():
             {"foo": [{"foo": "bar", "foobar": 1}, {"foo": "baz", "foobaz": 2}]},
             None,
         )
+
+    @pytest.mark.asyncio
+    async def cancel_on_exception():
+        barrier = Barrier(2)
+        completed = False
+
+        async def succeed(*_args):
+            nonlocal completed
+            await barrier.wait()
+            completed = True  # pragma: no cover
+
+        async def fail(*_args):
+            raise RuntimeError("Oops")
+
+        schema = GraphQLSchema(
+            GraphQLObjectType(
+                "Query",
+                {
+                    "foo": GraphQLField(GraphQLNonNull(GraphQLBoolean), resolve=fail),
+                    "bar": GraphQLField(GraphQLBoolean, resolve=succeed),
+                },
+            )
+        )
+
+        ast = parse("{foo, bar}")
+
+        awaitable_result = execute(schema, ast)
+        assert isinstance(awaitable_result, Awaitable)
+        result = await asyncio.wait_for(awaitable_result, 1.0)
+
+        assert result == (
+            None,
+            [{"message": "Oops", "locations": [(1, 2)], "path": ["foo"]}],
+        )
+
+        # Unblock succeed() and check that it does not complete
+        await barrier.wait()
+        await asyncio.sleep(0)
+        assert not completed


Cito:
Thank you @mgorven - will have a look later this week or on the weekend.

src/graphql/execution/execute.py
```diff
@@ -459,12 +468,16 @@ async def get_results() -> dict[str, Any]:
...
-                    zip(
-                        awaitable_fields,
-                        await gather(*(results[field] for field in awaitable_fields)),
-                    )
-                )
+                tasks = {}
+                for field in awaitable_fields:
+                    tasks[create_task(results[field])] = field  # type: ignore[arg-type]
+
+                done, pending = await wait(tasks, return_when=FIRST_EXCEPTION)
```
mgorven: The other approach is to allow all fields to finish executing and then raise the exception.

src/graphql/execution/execute.py
```diff
@@ -459,12 +468,16 @@ async def get_results() -> dict[str, Any]:
...
+                for task in pending:
+                    task.cancel()
+
+                results.update((tasks[task], task.result()) for task in done)
```
mgorven: This maintains the previous behaviour of not setting any results if an exception is raised, but I could change this to set the results we do have and then raise the exception.

mgorven:
Check is failing because coverage is not 100%, but this is because `completed = True` in the test case is not executed which is exactly what we're testing for.

src/graphql/execution/execute.py
dacevedo12: Does this behave in-line with the spec regarding partial results?

https://spec.graphql.org/draft/#sec-Handling-Execution-Errors

mgorven: I think this is fine because the potentially completed fields are siblings of the field with an exception, and the exception is propagated to the parent which will either be `null` or propagate further up. In either case these siblings field won't be present in the response.

src/graphql/execution/execute.py
Cito: You could do it using gather() like this:

```python
                tasks = [create_task(results[field]) for field in awaitable_fields]
                try:
                    awaited_results = await gather(*tasks)
                except Exception:
                    # cancel all pending tasks before bubbling up the error
                    for task in tasks:
                        if not task.done():
                            task.cancel()
                    await gather(*tasks, return_exceptions=True)
                    raise
                results.update(
                    zip(
                        awaitable_fields,
                        awaited_results,
                    )
                )
```

src/graphql/execution/execute.py
```diff
@@ -459,12 +468,18 @@ async def get_results() -> dict[str, Any]:
...
             else:
-                results.update(
-                    zip(
-                        awaitable_fields,
-                        await gather(*(results[field] for field in awaitable_fields)),
-                    )
-                )
+                tasks = {}
+                for field in awaitable_fields:
+                    tasks[create_task(results[field])] = field  # type: ignore[arg-type]
```
Cito:
```suggestion
                tasks = {
                    create_task(results[field]): field for field in awaitable_fields
                }
```

src/graphql/execution/execute.py
Cito: Manually re-raising the exception shows more explicitly what happens in the error case.

But your approach is also fine. Maybe add a comment before the results.update() mentioning that potential exceptions will be raised at this point automatically (when task.result() is called).

tests/execution/test_parallel.py
```diff
@@ -193,3 +194,40 @@ async def is_type_of_baz(obj, *_args):
             {"foo": [{"foo": "bar", "foobar": 1}, {"foo": "baz", "foobaz": 2}]},
             None,
         )
+
+    @pytest.mark.asyncio
+    async def cancel_on_exception():
+        barrier = Barrier(2)
```
Cito, line 200: Nice test, thank you.

tests/execution/test_parallel.py
```diff
@@ -193,3 +194,40 @@ async def is_type_of_baz(obj, *_args):
...
+        )
+
+        ast = parse("{foo, bar}")
+
+        awaitable_result = execute(schema, ast)
+        assert isinstance(awaitable_result, Awaitable)
+        result = await asyncio.wait_for(awaitable_result, 1.0)
+
+        assert result.errors
+        assert not result.data
```
Cito:
```suggestion
        assert result == (
            None,
            [
                {
                    "message": "Oops",
                    "locations": [(1, 2)],
                    "path": ["foo"],
                }
            ],
        )
```

To make sure the exception bubbles up properly.

tests/execution/test_parallel.py
```diff
@@ -193,3 +194,40 @@ async def is_type_of_baz(obj, *_args):
...
+        barrier = Barrier(2)
+        completed = False
+
+        async def succeed(*_args):
+            nonlocal completed
+            await barrier.wait()
+            completed = True
+
+        async def fail(*_args):
+            raise Exception
```
Cito:
```suggestion
            raise RuntimeError("Oops")
```

tests/execution/test_parallel.py
```diff
@@ -193,3 +194,40 @@ async def is_type_of_baz(obj, *_args):
...
+
+    @pytest.mark.asyncio
+    async def cancel_on_exception():
+        barrier = Barrier(2)
+        completed = False
+
+        async def succeed(*_args):
+            nonlocal completed
+            await barrier.wait()
+            completed = True
```
Cito:
```suggestion
            completed = True  # pragma: no cover
```

To avoid the coverage error.

Cito (changes requested):
Thanks a lot for this contribution and sorry for not looking into it earlier.

Your approach looks good to me, just a few suggestions.

Also, I think we need to do something similar in other places where we gather results asynchronously, e.g. in `complete_list_value()`. Let me know if you want to add it to this PR or create a separate one, or whether I should work on this.

mgorven:
Thanks for the review, I've addressed the comments. You can take the other instances, unless you aren't able to get to that for a while in which case I can help.

Cito (approved):

Cito:
Ok, will try to do something similar for the other instances as well and then create a release in the coming days.

Cito:
Thank you again for your contribution to this project.

Cito:
This fix is now been released in v3.3.0a8, I also fixed the other similar instances and extracted the extra code into a utility function.

PR #240, opened 2025-08-21
Fix AttributeError on exception with field `source`

kimjh12:
trying to fix below error, which happens if the handler raises an custom exception with `source` field inside

```
            locations: list[SourceLocation] | None = [
>               source.get_location(pos) for pos in positions
                ^^^^^^^^^^^^^^^^^^^
            ]
E           AttributeError: 'str' object has no attribute 'get_location'

.tox/py312/lib/python3.12/site-packages/graphql/error/graphql_error.py:161: AttributeError
```

### DIFF
--- a/src/graphql/error/located_error.py
+++ b/src/graphql/error/located_error.py
@@ -6,6 +6,7 @@ from contextlib import suppress
 from typing import TYPE_CHECKING, Collection
 
 from ..pyutils import inspect
+from ..language.source import is_source, Source
 from .graphql_error import GraphQLError
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


Cito:
Thank you @kimjh12.

Some questions to help me understand this better:

- Can you see which code created the GraphQLError with source which was a str object? Was it your own code?
- Would it make sense to set source=Source(source) if isinstance(source, str) else None?
- Wouldn't it be better to put the test in `test_located_error`?

Cito:
Hi @kimjh12 I am planning to do some maintenance on GraphQL-core next week, so I'd appreciate your feedback to fix this properly.

kimjh12:
> * Can you see which code created the GraphQLError with source which was a str object? Was it your own code?

Yes, the issue was with my team's code. I updated the diff, thanks

Cito (approved):

Issue #166, opened 2022-04-03
Proposal: replace asyncio with anyio


valgarf:
Yesterday I added the pull request #165 to start a discussion on supporting the trio ecosystem, preferably using anyio. @Cito invited me to create an issue as a place for discussing the implications of such a step. 

**I hope to get some feedback from you wether you think supporting trio in graphql-core is a good idea.** My reasoning and approach for the pull request is outlined below

# The basics

## What is structured concurrency and trio?

This is best answered by the trio project, I can recommend reading their [documentation](https://trio.readthedocs.io/en/stable/), especially their [design principles](https://trio.readthedocs.io/en/stable/design.html). There are also a few blog posts that shed light on the reasoning behind trio, e.g. [structured concurrency](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/) and [cancellation](https://vorpus.org/blog/timeouts-and-cancellation-for-humans/).

Basically, trio implements an event loop for asynchronous code execution, just like the asyncio standard library. However, there are some important differences. The following two have probably the most impact:

1.) You cannot start async tasks whenever you want. All tasks have a parent, if you want to start multiple coroutines, you need a `Nursery`. The nursery guarantees that all tasks are done when exiting its scope. This makes it much easier to reason about async code: If you await an async function, there will be no background tasks running when the function returns. Only exception: passing in a nursery explicitely to start the tasks. Then the tasks may run until the end of the nursery scope.

2.) Level-based cancellation. Cancellation is done via scopes. Once a scope is cancelled, all its children scopes are cancelled as well. It also stays cancelled: After cancellation, any checkpoint (see trio docs for details, basically any call into trio like `trio.sleep`) raises `Cancelled`. This makes it easier to get cancellation right. Asyncio only ever sends one `CancelledError`. It is often caught with good intentions but not raised again in some circumstances. 

These two changes alone make it much easier to write correct asynchronous code.

## What is anyio?

The [anyio project](https://anyio.readthedocs.io/en/stable/) Allows to use the same structured concurrency and cancellation behavior as trio (trio's `Nursery` is a `TaskGroup`, etc.). However, it can run on two different backends: trio and standard asyncio.

This makes anyio ideal for two use cases: 
  1.) Iteratively replacing asyncio code with anyio code. At any point the code should still work, even if only partially replaced. 
  2.) Writing libraries that support trio and asyncio

# Why support trio in graphql-core?

[...]

PR #242, opened 2025-09-15
Use AnyIO instead of pytest-asyncio for running async tests

agronholm:
This is the first step to implement #166.


agronholm:
The linting failure is not due to my changes, it exists in `main` already.

Cito:
Thanks for the contribution, @agronholm. Will review later. Never mind any linting issues, will fix them.

Cito (approved):
Looks good, thank you.

Will remove pytest-asyncio from pyproject.toml after merging.

agronholm:
> Looks good, thank you.
> 
> Will fix the mypy error and remove pytest-asyncio from pyproject.toml after merging.

Ah, not removing it from `pyproject.toml` was an oversight on my part. Have you considered using standard dependency groups? Tox supports them, allowing you to reduce duplication in the dependency specifications.

Cito:
> Have you considered using standard dependency groups? Tox supports them

I'm actually defining them, but not using them in tox. Might improve that, good idea. Wanted to improve the dependency management anyway, probably switching from poetry to uv.

agronholm:
> > Have you considered using standard dependency groups? Tox supports them
> 
> I'm actually defining them, but not using them in tox. Might improve that, good idea. Wanted to improve the dependency management anyway, probably switching from poetry to uv.

You have dependency groups but they're using the Poetry specific syntax. Recent releases of Poetry should support the standard syntax under the `dependency-groups` key.

Cito:
Thanks, made a note to change that. But as already mentioned, might completely move away from Poetry anyway.

Cito:
@agronholm The project is now standard dependency groups, and uv instead of poetry, resulting in less duplication and more speed. Thanks for pushing me.

PR #245, opened 2025-10-24
fix: don't loose extensions of enum values when sorting schemas

erikwrede:
When sorting a schema lexicographically, the enum values lost their extensions.


Cito (approved):

Cito:
Thank you @erikwrede

Cito:
Actually input values and the schema itself had the same problem, fixed that already, and backported it as well.

erikwrede:
@Cito still meant to check for that, great you found it first :) thanks for the quick merge

Cito:
Just for the records: We overlooked that the same problem existed for the deprecation reason and out name, but this has been fixed now as well with #264.

PR #259, opened 2026-02-28
Switch to typing_extensions.Self for __new__ implementations

bigfootjon:
I noticed that when running with https://pyrefly.org/ I get type errors on things like:

```python
from graphql import GraphQLEnumType

def get_enum() -> GraphQLEnumType:
  return GraphQLEnumType()
```

This occurs because the `__new__` function for `GraphQLNamedType` returns `GraphQLNamedType` (i.e. that *exact* class) instead of `Self` which automatically infers the actual class (e.g. `GraphQLEnumType` in this case).

I used the `typing_extensions` version because AIUI this project supports versions of python from before `Self` was introduced, and I also fixed the other `__new__` function that returns a type while I was fixing the problem I was seeing.

### DIFF
--- a/src/graphql/pyutils/undefined.py
+++ b/src/graphql/pyutils/undefined.py
@@ -1,9 +1,15 @@
 """The Undefined value"""
 
 from __future__ import annotations
-from typing_extensions import Self
 
 import warnings
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    try:
+        from typing import Self
+    except ImportError:  # Python < 3.11
+        from typing_extensions import Self
 
 __all__ = ["Undefined", "UndefinedType"]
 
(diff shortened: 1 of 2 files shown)


leszekhanusz:
In gql, we are importing `Self` like this:

```python
import sys
if sys.version_info >= (3, 11):
    from typing import Self  # pragma: no cover
else:
    from typing_extensions import Self  # pragma: no cover
```

I think it makes it easier for the maintainer to know when we can avoid using `typing_extensions` when we remove the support for an old Python version.

Cito (changes requested):
Thank you @bigfootjon, this makes sense.

As @leszekhanusz said, this should be only imported conditionally, and the import should go to the TYPE_CHECKING block.

Usually I don't even check the sys.version, but just try to import from typing and then from type extensions if that fails - like [here](https://github.com/graphql-python/graphql-core/blob/81ef8aa362c65eef1cd6e9beb184f2c259e1eb0a/src/graphql/execution/types.py#L23). But I add the Python version as comment, so that it can be searched and removed when no longer needed.

Cito (approved):

Issue #140, opened 2021-09-22
Camel to snake case conversion doesn't add an underscore after a digit


Matt343:
The `pyutils.camel_to_snake` function does not add an underscore if a capital letter occurs after a non-letter character. For example, `came_to_snake("python2Thing")` returns `python2thing` instead of `python2_thing`. This seems counterintuitive, as it changes the case of the `T` without adding an underscore as usual.

Cito:
Thanks for reporting. I agree, `python2_thing` would be the expected conversion. It's also what I get when I convert it with PyCharm or Visual Studio Code.

The pyutils function currently uses the same method as [python-string-utils](https://github.com/daveoncode/python-string-utils), which has this "bug". I'll see if I can find a better way when I find some spare time. If you know any better method, please let me know.

gosella:
In the past, I used this conversion function with great success:
```
>>> import re

>>> CAMEL_CASE_TO_SNAKE_CASE_RE = re.compile('((?<=[a-z0-9])[A-Z]|(?!^)[A-Z](?=[a-z]))')

>>> def camel_to_snake(a_string):
...     return CAMEL_CASE_TO_SNAKE_CASE_RE.sub(r'_\1', a_string).lower()

>>> camel_to_snake('python2Thing')
'python2_thing'
```
Credits to https://stackoverflow.com/a/12867228 (last given example)

Cito:
Thanks @gosella. I solved it in a similar way by making the minimal necessary change to the existing regex so that the "python2Thing" works as expecetd. If there are edge cases where your regex works better let me know, we can always add more tests and improve.

PR #261, opened 2026-03-07
camel_to_snake adds underscore after digits

BlackBeardsoul:
Fixed regex to add underscore between digit sequences and uppercase letters. Example: camel_to_snake('python2Thing') now returns 'python2_thing' instead of 'python2thing'.


Cito (commented):
The issue #140 had already been solved in 0ed4383. I have now added an extra test case to confirm that.

PR #265, opened 2026-05-13
Add protocol for hashable async iterators in `execution/execute.py`

jonathandung:
Some parameters typed as async iterators in the above file are added to sets, which require them to be hashable as well, so a protocol was added to type that.

Parent: [python/typeshed#15754](https://github.com/python/typeshed/pull/15754)


Cito:
Many thanks for pointing out this typing issue. The canceled iterators set was actually unused (left over after a refactoring). I removed it, so we don't need to fix anything here any more.

PR #273, opened 2026-07-13
Fix "Invalid state while adding deferred fragment node" when a @defer'd field errors

### LINKED ISSUE #271: Incremental delivery: RuntimeError "Invalid state while adding deferred fragment node" when a @defer'd field errors
Hi there,

I've been working to resolve a production error we've logged a few thousand times in the last 48 hours - below is the diagnosis and minimal reproduction case that I homed in on with the help of Claude code.

I raised a fix PR - #273 .
 
- Ben (Mixcloud)

---

## Summary

`experimental_execute_incrementally` raises an internal `RuntimeError: Invalid state while adding deferred fragment node.` on a **spec-valid** query when **two sibling top-level `@defer` fragments — each containing a field that errors and its own nested `@defer`** — are executed with async resolvers. The crash is in the `# pragma: no cover` branch of `IncrementalGraph._add_deferred_fragment_node`.

We hit this **in production** (Strawberry over ASGI, streaming `@defer` as `multipart/mixed`): ~5,700 occurrences / 48h across ~4,100 users, firing mid-stream after the HTTP 200 status line (so clients get a truncated incremental response). It reproduces on **3.3.0a12 through 3.3.0rc0** (current tip).

## Environment

- graphql-core `3.3.0rc0` (also `3.3.0a12`)
- Python 3.14 (also seen on 3.13)

## Query (spec-valid — passes `graphql.validate`)

```graphql
query {
  obj {
    ... @defer(label: "a") { boom   child { ... @defer(label: "b") { fast } } }
    ... @defer(label: "c") { boomNN child { ... @defer(label: "d") { slow } } }
  }
}
```

`boom` / `boomNN` are async resolvers that raise; `fast` / `slow` / `child` are ordinary async resolvers. Small random delays perturb execution-group completion ordering (the bug is a race, so the repro loops).

## Stacktrace

```
File ".../graphql/execution/incremental_publisher.py", line 248, in _handle_completed_execution_group
    self._incremental_graph.add_completed_successful_execution_group(completed_execution_group)
File ".../graphql/execution/incremental_graph.py", line 86, in add_completed_successful_execution_group
    self._add_incremental_data_records(incremental_records, deferred_records)
File ".../graphql/execution/incremental_graph.py", line 192, in _add_incremental_data_records
    self._add_deferred_fragment_node(deferred_fragment_record, initial_result_children)
File ".../graphql/execution/incremental_graph.py", line 266, in _add_deferred_fragment_node
    self._add_deferred_fragment_node(parent, initial_result_children)   # recurse to parent
File ".../graphql/execution/incremental_graph.py", line 262, in _add_deferred_fragment_node
    raise RuntimeError(msg)
RuntimeError: Invalid state while adding deferred fragment node.
```

## Minimal trigger (each part is required)

| Variant | Reproduces? |
| --- | --- |
| Two sibling top-level `@defer`, **each** with an erroring field **and** a distinct nested `@defer` | **yes (~1 in 4 runs)** |
| Only one sibling has a nested `@defer` | no |
| Single deferred fragment (not two siblings) | no |
| The two nested `@defer`s share a label (dedupe to one record) | no |
| No erroring field | no |

## Mechanism

When a deferred fragment's field errors, that top-level fragment's execution group fails and the fragment is removed from `_root_nodes`. A **sibling** group completing afterwards (`add_completed_successful_execution_group → _add_incremental_data_records`) discovers its nested `@defer` and, via `_add_deferred_fragment_node`, recurses up to the now-removed parentless ancestor. With `parent is None` and `initial_result_children is None`, it hits the `# pragma: no cover` branch and raises:

```python
def _add_deferred_fragment_node(self, deferred_fragment_record, initial_result_children=None):
    if deferred_fragment_record in self._root_nodes:
        return
    parent = deferred_fragment_record.parent
    if parent is None:
        if initial_result_children is None:  # pragma: no cover
            msg = "Invalid state while adding deferred fragment node."
            raise RuntimeError(msg)          # <-- reached via the resolver-error race
        ...
```

## Suggested direction

Reaching a parentless record that is no longer in `_root_nodes` here looks tolerable rather than fatal — the errored parent's subtree isn't being delivered anyway — so returning/no-op instead of raising would mirror the `future.cancelled()` guard recently added to `_enqueue` ("defensive guard against a race with a stopping consumer"). Happy to open a PR if that direction sounds right.

### Self-contained reproduction (graphql-core only)

```python
import asyncio
import random

from graphql import (
    GraphQLDeferDirective,
    GraphQLField,
    GraphQLNonNull,
    GraphQLObjectType,
    GraphQLSchema,
    GraphQLString,
    parse,
    specified_directives,
    validate,
)
from graphql.execution import (
    ExperimentalIncrementalExecutionResults,
    experimental_execute_incrementally,
)
from graphql.pyutils import is_awaitable

MAX_ITERS = 500


async def value(v: str) -> str:
    await asyncio.sleep(random.uniform(0, 0.002))  # perturb completion ordering
    return v


async def boom(_source, _info) -> str:
    await asyncio.sleep(random.uniform(0, 0.002))
    raise RuntimeError("resolver error")


async def obj(_source, _info) -> dict:
    await asyncio.sleep(random.uniform(0, 0.001))
    return {}


OBJ = GraphQLObjectType(
    "Obj",
    lambda: {
        "fast": GraphQLField(GraphQLString, resolve=lambda *a: value("fast")),
        "slow": GraphQLField(GraphQLString, resolve=lambda *a: value("slow")),
        "boom": GraphQLField(GraphQLString, resolve=boom),
        "boomNN": GraphQLField(GraphQLNonNull(GraphQLString), resolve=boom),
        "child": GraphQLField(OBJ, resolve=obj),
    },
)
schema = GraphQLSchema(
    GraphQLObjectType("Query", {"obj": GraphQLField(OBJ, resolve=obj)}),
    directives=[*specified_directives, GraphQLDeferDirective],
)

QUERY = """
query {
  obj {
    ... @defer(label: "a") { boom   child { ... @defer(label: "b") { fast } } }
    ... @defer(label: "c") { boomNN child { ... @defer(label: "d") { slow } } }
  }
}
"""


async def run_once() -> None:
    result = experimental_execute_incrementally(schema, parse(QUERY), root_value={})
    if is_awaitable(result):
        result = await result
    if isinstance(result, ExperimentalIncrementalExecutionResults):
        async for _patch in result.subsequent_results:
            pass


async def main() -> None:
    import graphql

    print("graphql-core:", graphql.__version__)
    assert not validate(schema, parse(QUERY)), "query should be valid"
    for i in range(1, MAX_ITERS + 1):
        try:
            await run_once()
        except RuntimeError as exc:
            if "Invalid state while adding deferred fragment node" in str(exc):
                print(f"Reproduced on iteration {i}")
                raise
            raise
    print(f"Did not reproduce in {MAX_ITERS} iterations (try re-running).")


asyncio.run(main())
```

---

_This crash, and its minimal trigger, were discovered and confirmed by fuzzing `@defer`/`@stream` query shapes with erroring async resolvers._

ben-xo:
Fixes #271 (which is an issue we were seeing in production).

## Problem

`IncrementalGraph._add_deferred_fragment_node` raises `RuntimeError: Invalid state while adding deferred fragment node.` while executing a **spec-valid** query, when two sibling top-level `@defer` fragments — each resolving a field that errors and each containing its own nested `@defer` — are executed with async resolvers.

We hit this in production (Strawberry over ASGI, streaming `@defer` as `multipart/mixed`) at ~5.7k occurrences / 48h.

## Cause

When a deferred field errors, that fragment's execution group fails and the (parentless, root) deferred fragment is removed from `_root_nodes`. A sibling execution group completing afterwards then discovers its nested `@defer` and, in `add_completed_successful_execution_group → _add_incremental_data_records → _add_deferred_fragment_node`, recurses up to the now-removed parentless fragment — reaching the `# pragma: no cover` branch and raising.

## Fix

Return instead of raising when the recursion reaches a parentless fragment that is no longer a root node: its subtree is no longer being delivered, so there is nothing to attach the record to. This mirrors the `future.cancelled()` guard in `_enqueue` ("defensive guard against a race with a stopping consumer").

## Test

Adds a regression test in `tests/execution/test_defer.py` that reproduces the crash before this change (raising the exact `RuntimeError`) and passes after it. The full `tests/execution/` suite passes.

The crash and its minimal trigger were discovered and reduced via fuzzing of `@defer`/`@stream` shapes with erroring async resolvers.

### DIFF
--- a/src/graphql/execution/incremental_graph.py
+++ b/src/graphql/execution/incremental_graph.py
@@ -257,9 +257,16 @@ class IncrementalGraph:
             return
         parent = deferred_fragment_record.parent
         if parent is None:
-            if initial_result_children is None:  # pragma: no cover
-                msg = "Invalid state while adding deferred fragment node."
-                raise RuntimeError(msg)
+            if initial_result_children is None:
+                # The fragment has already been removed from the root nodes
+                # (e.g. its execution group failed on a resolver error) while a
+                # sibling execution group that references its subtree is still
+                # completing. There is nothing to attach the record to and its
+                # subtree is no longer being delivered, so drop it rather than
+                # treating this as an invalid state. Mirrors the
+                # `future.cancelled()` guard in `_enqueue` — a race with a
+                # stopping/removed consumer.
+                return
             initial_result_children[deferred_fragment_record] = None
             return
         parent.children[deferred_fragment_record] = None
--- a/tests/execution/test_defer.py
+++ b/tests/execution/test_defer.py
@@ -3163,3 +3163,78 @@ def describe_execute_defer_directive():
         with pytest.raises(AbortError, match="This operation was aborted"):
             await anext_task
         assert items_source.aclose_finished
+
+
+def describe_defer_directive_with_errors_and_nested_defer():
+    """Regression tests for graphql-python/graphql-core#271."""
+
+    async def does_not_reach_invalid_state_when_sibling_fragments_error():
+        """Sibling erroring ``@defer`` fragments with nested ``@defer`` must not crash.
+
+        Two sibling top-level ``@defer`` fragments that each resolve a distinct
+        erroring field and each contain a nested ``@defer`` used to crash the
+        incremental graph with "Invalid state while adding deferred fragment
+        node": a failing
+        execution group removes its parentless (root) deferred fragment, and a
+        sibling execution group completing afterwards then recursed up into the
+        removed record and hit the invariant. See issue #271.
+        """
+
+        async def resolve_fast(_source, _info) -> str:
+            await sleep(0)
+            return "fast"
+
+        def resolve_error(_source, _info) -> str:
+            raise RuntimeError("bad")
+
+        async def resolve_obj(_source, _info) -> dict:
+            await sleep(0)
+            return {}
+
+        obj_type: GraphQLObjectType = GraphQLObjectType(
+            "Obj",
+            lambda: {
+                "fast": GraphQLField(GraphQLString, resolve=resolve_fast),
+                "boom": GraphQLField(GraphQLString, resolve=resolve_error),
+                "boomNonNull": GraphQLField(
+                    GraphQLNonNull(GraphQLString), resolve=resolve_error
+                ),
+                "child": GraphQLField(obj_type, resolve=resolve_obj),
+            },
+        )
+        local_schema = GraphQLSchema(
+            GraphQLObjectType(
+                "Query", {"obj": GraphQLField(obj_type, resolve=resolve_obj)}
+            )
+        )
+        # The two siblings must error on DISTINCT fields (`boom` vs `boomNonNull`):
+        # erroring on the same field in both does not reproduce the crash. The
+        # nested `@defer` selections may be identical.
+        document = parse(
+            """
+            query {
+              obj {
+                ... @defer(label: "a") {
+                  boom
+                  child { ... @defer(label: "b") { fast } }
+                }
+                ... @defer(label: "c") {
+                  boomNonNull
+                  child { ... @defer(label: "d") { fast } }
+                }
+              }
+            }
+            """
+        )
+
+        for early_execution in (False, True):
+            result = experimental_execute_incrementally(
+                local_schema, document, {}, enable_early_execution=early_execution
+            )
+            assert is_awaitable(result)
+            result = await result
+            assert isinstance(result, ExperimentalIncrementalExecutionResults)
+            # Draining the stream must terminate without raising
+            # "Invalid state while adding deferred fragment node".
+            async for _patch in result.subsequent_results:
+                pass


Cito:
Hi @ben-xo. Thanks a lot for the analysis and PR with regression test which all look solid to me at a first glance. Will have a deeper look later.

This issue occurs in a part of the code which has since been rewritten using a WorkQueue/IncrementalExecutor approach. I did not fully port this since it is heavily based on promises and hard to translate to the world of asyncio. However, this makes me think if I should still try to port it to Python somehow so that we stay aligned with upstream and avoid such problems.

ben-xo:
I noticed the build failed on mypy - double checked - I don't think it's my work, as the failures are in pre-existing files. (Just wanted to be diligent). Seems like there was a new release of mypy yesterday.

ben-xo:
I've raised #274 for the mypy errors against mypy 2.3.0 (unrelated to this PR, but hope it helps)

Cito:
Yes, the mypy errors are caused by the mypy update. Will fix via your other. Thanks.

